PR #77, opened 2020-08-16
Hide Accounting Dimension and add filter in project

abrarpv97:
@ks093 
1.created new js to Remove Mandatory of "Cost Center" and Hide "Accounting Dimension" Section in Sales Invoice child table
2.Set Customer filter in project in Sales Invoice
3.Added doctype link in hooks.py

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -46,6 +46,7 @@ doctype_js = {
 	"Shift Type" : "public/js/doctype_js/shift_type.js",
 	"Project": "public/js/doctype_js/project.js",
 	"Notification Log": "public/js/doctype_js/notification_log.js",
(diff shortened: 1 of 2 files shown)


abrarpv97:
3.firstly i had checked with my own data so then default cost_center from
company will populated on saving sales invoice .
now i have checked by taking backup of one-fm and it has no default
cost_center .creating a sales invoice without cost_center in sales invoice
item table is not possible

abrarpv97:
@ks093 
3.firstly i had checked with my own data so then default cost_center from company will populated on saving sales invoice .
now i have checked by taking backup of one-fm and it has no default cost_center .creating a sales invoice without cost_center in sales invoice item table is not possible

abrarpv97:
i have created new pull request

PR #82, opened 2020-08-24
[Fix]: Custom field changes

pjamsheer:
- fix: Job Offer Custom field from fixtures and added in one_fm/custom/job_offer.json
- fix: Job Opening and Designation Skill Custom field from fixtures and added in one_fm/custom/job_offer.json and one_fm/custom/designation_skill.json


ks093 (changes requested):
@pjamsheer the custom_field.json file still contains project_full_name and project_sites fields. Please remove that. Other than that, looks fine.

ks093 (approved):
Looks fine.

PR #234, opened 2021-01-18
site wise invoice /Pifss /MOI

abrarpv97:
@ks093 
please check 
1. added site field in timesheet detail
2. added site field in sales invoice item
3. setting value to site in timesheet auto creation
4. site wise invoice in sales invoice auto creation
5. checking po attatch field before submitting invoice
6. added po custom field in sales invoice


ks093 (changes requested):
@abrarpv97 Please break the create_sales_invoice function into smaller functions. It is almost 380 lines long. 
Also there is an empty return statement at the end.

abrarpv97:
@ks093 
i have made requested changes

abrarpv97:
@ks093 
please check new commits for pifss and MOI

abrarpv97:
@ks093 
i have made changes as you requested,

PR #290, opened 2021-03-29
fix:Modify employee fields

Amna-onefm:
Add custom field for filtering the designation of the employee based on the PAM designation selection.


pjamsheer:
@Amna-onefm  There are lot of commented codes, please do a code clean up.

Amna-onefm:
Ok I will.

PR #322, opened 2021-05-08
Changes in Item doc.....Changes in RFM and RFP

osamaasim:


pjamsheer (changes requested):
Item custom fields:
- There is a section break called Inventory and it has Shelf Life In Days and End of Life - make use of it instead of start and end date in life time.
- We have Brand itself in Item no need of new doctype Item Brand
- Why Item UOM field, we have Stock UOM, Sales UOM, Purchase UOM in Item
- We have a Supplier Details Section in Item, There we can select as many supplier we need. No need to set a new field for that

Item description setup:
- Need to review once

osamaasim:
Requested changes implemented.

PR #331, opened 2021-05-21
changes in update password

SalShai:


ks093 (changes requested):
@SalShai Please add the changes for Login and forgot password functions also.

SalShai:
Okay

Salwa Shaikh

PR #337, opened 2021-05-28
Fix: priority level and empty designation

Ehtesham599:
- Changed priority level starting from 5 to starting from 10.
- Temporary fix for employee designation: Fetch designation directly from employee doctype instead of employee skill map.
- Update employee assignment in employee doctype upon change of shift.


ks093 (changes requested):
@Ehtesham599 
Instead of doing frappe.get_doc for employee designation in post_allocation_plan.py,  
use frappe.db.get_value.
It's better performance wise.

Ehtesham599:
@ks093 
Noted and made the changes.

PR #351, opened 2021-06-07
feature:Changes in RFM

osamaasim:


pjamsheer:
@osamaasim while setting erf items to RFM, you should capture item UOM and item code if any in erf to RFM.

pjamsheer (approved):

PR #355, opened 2021-06-09
Fix: Changes on wp,moi,mi,paci records

Amna-onefm:


pjamsheer:
@Amna-onefm  instead of setting field_type as Read Only, use Data and to achieve read only set read only. Field type customisation is not possible, but we can change the read only property by customisation and custom script.

PR #363, opened 2021-06-17
Fix: Added role permissions, updated post allocation plan and included "service type" in operations shift.

Ehtesham599:
- Added Operations Manager role to Certification and License doctypes.

### DIFF
--- a/one_fm/api/doc_events.py
+++ b/one_fm/api/doc_events.py
@@ -354,12 +354,20 @@ def validate_certifications_and_licenses(doc, method):
 
 	if len(certifications) > 0:
 		for certification in certifications:
-			if(getdate(certification.expiry_date) <= getdate()):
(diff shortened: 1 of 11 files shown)


Ehtesham599:
- Added a new field "Service Type" in the Operations Shift doctype and linked it with the Service Type doctype.
- Changed autonaming format of the Operations Shift doctype and checked "Allow rename" option for the same.
- Added validation to check if updated document matches the naming format specified in the autonaming function. If not, rename the updated document.

pjamsheer:
@Ehtesham599 
-[ ] Please check line number line number 17 one_fm/operations/doctype/operations_shift/operations_shift.py
-[ ] Why rename in validate of Operations Shift

pjamsheer (approved):

PR #384, opened 2021-07-01
Training program certification process

Ehtesham599:
- Created Training Program Certificate doctype and a script report for the same.
- Changed link in employee certification and post certification to Training Program doctype.
- Added controller to create a Training Program Certificate for an employee based on Training event's has certificate condition and Training Result's pass/fail criteria and update Employee Skill Map for the certificate obtained.


pjamsheer:
@Ehtesham599 

- [ ] Instead of having fieldtype Read Only, put fieldtype as Data and mark check for read_only, since we can set read_only true and false from the custom script and cutomize form if we need, but changing field type is not possible from front end.(Training Program Certification Doctype)
- [ ] In training Program Certificate Report - You can extend filters in js file with fieldtype as 'Link' and add options to 'Training Program'
- [ ] In training Program Certificate Report - In py file, line number 29 you can combine the if statements with and.

pjamsheer (approved):

Ehtesham599:
@pjamsheer Thank you for your feedback. Made the required changes.
Kind request to review them.

pjamsheer (approved):

PR #395, opened 2021-07-07
Changes:return site and location in same function

SalShai:


pjamsheer:
@SalShai 
if there is nothing in location, what will be location[0] ?

SalShai:
If these is nothing is location, it'll return None.

SalShai:
@pjamsheer I'm sorry, I didn't think of that condition. I have made the required changes. Kindly, Check.

pjamsheer:
@SalShai Please do if check before you assigning location[0] to some field, since if location is not there it will through list index out of range

pjamsheer (approved):

PR #400, opened 2021-07-12
Dev1

Amna-onefm:


pjamsheer:
@Amna-onefm There are two Transfer Paper Doctype found. One in Hiring Module and the other in One FM module.

Amna-onefm:
@pjamsheer Please just merge the one in the hiring module as the one in One FM pushed by mistake and wasn't able to checkout the file after adding it into git.

PR #411, opened 2021-07-15
Dev2

Amna-onefm:
- Adding fields in pam authorized signatory.
- Removing fields in job applicant.
- Linking selection of pam file with the authorized signatory of the pam file.


Amna-onefm:
@pjamsheer, Kindly can you review the PR and merge it.

PR #483, opened 2021-08-24
Fix: Payroll Entry Generation

SalShai:


pjamsheer:
@SalShai Account selection should be from some Configuration, since there will be chances to have changes in the words in the Account Head

SalShai:
@pjamsheer  It's a mandatory field. Couldn't submit the form without it. Can you please suggest me another way?

pjamsheer (approved):

PR #503, opened 2021-09-01
Update: arabic field for gender doctype

Amna-Alshawa:


pjamsheer:
@Amna-onefm field name religion_english can be religion only

Amna-Alshawa:
@pjamsheer, does it matter?

pjamsheer (approved):

PR #507, opened 2021-09-01
Fix: allow on submit

Amna-Alshawa:


pjamsheer:
@Amna-onefm ERF is a Note ERPNext Code Doctype, It is One FM Doctype, Please update in the DOctype itself and Update

pjamsheer (approved):

PR #579, opened 2021-10-06
Push Notification 

SalShai:


pjamsheer:
@SalShai There are lots of print, in server we dont need any print.

SalShai:
@pjamsheer I'll remove them.

SalShai:
@pjamsheer Could you check now??

pjamsheer:
@SalShai one_fm/api/tasks.py in this file the notification messgae creation function can be optimised and reused for the two conditions

SalShai:
okay. I'll check.

pjamsheer:
@SalShai 
instead of `if status is None: or assigning status = None` use if status itself and status = False

SalShai:
> @SalShai instead of `if status is None: or assigning status = None` use if status itself and status = False

@pjamsheer I tried this method before, but it did not work.

pjamsheer:
@SalShai We can have a review meeting

SalShai:
@pjamsheer Yes. sure.
When can I schedule it?

SalShai:
@pjamsheer I have made the changes that was suggested by you. and have tested it as well. Kindly, Review it.

pjamsheer (approved):

PR #585, opened 2021-10-12
HotFix: nonetype value

Amna-Alshawa:


pjamsheer:
@Amna-onefm  `one_fm_number_of_kids` should be `int`, otherwise user can enter any data value ?!
If the field type is `int` 
then
instead of `if doc.one_fm_number_of_kids > 0:` you can put 
`if doc.one_fm_number_of_kids and doc.one_fm_number_of_kids > 0:` and update the PR

pjamsheer (approved):

PR #599, opened 2021-10-24
Fix: fetch employee Name/ID to fix list index out of range

SalShai:


Amna-Alshawa (approved):
@pjamsheer , PR has been reviewed and approved.

PR #606, opened 2021-10-26
Fixed working days count in roster

Ehtesham599:
- Fixed attendance 'present' value in condition check


SalShai (approved):
@pjamsheer , this has been reviewed and Approved. You can merge it.

PR #615, opened 2021-10-27
#179977547: Details of a Shift Permission

SalShai:


Ehtesham599 (changes requested):
@SalShai Exceptions must be returned by the method instead of printing them.
Example: `return frappe.utils.response.report_error(e.http_status_code)`

SalShai:
okay👍🏼

SalShai:
@Ehtesham599, I might close this and create a new one.

PR #630, opened 2021-11-01
[Fix] Annual Leave and Amount Calculation Report

pjamsheer:


Amna-Alshawa:
**@pjamsheer**
- [x] Replace ToDo in line 94 with a normal standard comment as it is no more in the todo list. 
- [x] Salary_per_day in line 74 is for 30 days month, what about months with 28 & 31 days! - # assuming that 30 days of month (asper the requirement) - will be do in next iterations
- [x] Line 97 what if relieving_date is NoneType, Did you set it as required somewhere? - # in L#90 it is checking that `if employee.relieving_date` and  employee.relieving_date is date field

pjamsheer:
@Amna-onefm Please check

Amna-Alshawa (approved):
@pjamsheer , PR has been approved.

PR #631, opened 2021-11-01
Roster Report

Ehtesham599:
- Implemented a function that generates roster report in the background and send it via email to operation managers.
- Displayed report data in an HTML tabular format.
- Added roster report generation to monthly scheduler event.


SalShai (approved):
@pjamsheer, Approved👍🏼

PR #642, opened 2021-11-07
Post Type Actions [Enhancement and bug fixes]

Ehtesham599:
## Feature description
A function that creates a Roster Post Actions document specifying the supervisor in charge, post types not filled under the same supervisor and the action to be taken, fetching post types not filled in a date range of 14 days (2 weeks).

For a span of two weeks, all post types that are not filled are fetched. The supervisors that are supervising the shifts that the post types are scheduled in are then grouped together with these post types.
The supervisors are then notified, issuing them actions to fill the post type in their shift.

## Solution description
First, the post types that are not filled are fetched by making comparisons in post schedules and employee schedules.
Then, the supervisors and the post types scheduled in their shifts are obtained.
Lastly, for each supervisor, issue a Roster Post action listing the post types to be filled.
The Roster Post Actions upon creation automatically sends an email notification to the supervisor, providing a link to the document.

## Areas affected and ensured
Nothing affected and ensured.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/one_fm/doctype/post_types_not_filled/post_types_not_filled.json
+++ b/one_fm/one_fm/doctype/post_types_not_filled/post_types_not_filled.json
@@ -1,0 +1,33 @@
+{
+ "actions": [],
+ "allow_rename": 1,
+ "creation": "2021-11-07 10:47:12.278579",
+ "doctype": "DocType",
+ "editable_grid": 1,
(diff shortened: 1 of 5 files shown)


Amna-Alshawa (approved):
@Ehtesham599 , Perfect.

PR #643, opened 2021-11-08
[Fix] Employee Annual Leave Calculation

pjamsheer:
## Feature description
To show all employees' annual paid leave balances, so that I am able to process annual leaves and termination

## Solution description
Report called Employee Annual Leave Calculation

No.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
Approved👍🏼 @pjamsheer

PR #645, opened 2021-11-08
Update: add filter in checkin and checkout notifications

Amna-Alshawa:
## Feature description
Adding a filter to eliminate employee who has accepted shift permission record from receiving check-in and check-out notifications.

## Solution description
Employees won't receive checkout and check-in notifications when accepted Leave Early and Arrive late shift permissions are present.

No

  - [x] Chrome
  - [ ] Safari


Ehtesham599 (approved):
LGTM!

PR #666, opened 2021-11-10
#179977491

SalShai:
## Feature description
push_notification Function to send push notifications to a group of devices. Here, we use the 'firebase admin' library to send the message.

## Solution description
The Custom Notification has 3 buttons that are enabled for different scenarios. For example, check-in reminder of head office needs 2 buttons, and others only need one.  The notification data carries the command to enable the buttons in need.

Final Reminder Check-in(Head Office):
Screenshot_20211107-094929 

Final Reminder Check-in:
Screenshot_20211107-094929 

Final Reminder Check-out:
Screenshot_20211107-094929 

## Areas affected and ensured
- Final Reminder

No.

  - [x] Chrome
  - [ ] Safari

### DIFF
--- a/one_fm/api/tasks.py
+++ b/one_fm/api/tasks.py
@@ -61,7 +61,7 @@ def final_reminder():
 		# Employee won't receive checkin notification when accepted Arrive Late shift permission is present
 		if strfdelta(shift.start_time, '%H:%M:%S') == cstr((get_datetime(now_time) - timedelta(minutes=cint(shift.notification_reminder_after_shift_start))).time()):
 			recipients = frappe.db.sql("""
-				SELECT DISTINCT emp.user_id, emp.name, tSA.shift FROM `tabShift Assignment` tSA, `tabEmployee` emp
+				SELECT DISTINCT emp.user_id, emp.name FROM `tabShift Assignment` tSA, `tabEmployee` emp
 				WHERE
 			  		tSA.employee=emp.name 
 				AND tSA.start_date="{date}"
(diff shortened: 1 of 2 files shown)


one_fm/api/tasks.py
```diff
@@ -121,17 +130,23 @@ def notify(recipients,log_type):
 	checkout_subject = _("Final Reminder: Please checkout in the next five minutes.")
 	checkout_message = _("""<a class="btn btn-danger" href="/desk#face-recognition">Check Out</a>""")
 	Notification_title = "Final Reminder"
-	user_id = []
-	employee_id = []
+	Notification_body = "Please checkin in the next five minutes."
+	
+	#eg: recipient: ['s.shaikh@armor-services.com', 'HR-EMP-00001', 'Develope-Head Office-Morning-1']
 	for recipient in recipients:
```
Ehtesham599, line 139: I think you can use recipient as an object instead of a list with keys: email_id, employee, shift.
That would make more sense.

one_fm/api/tasks.py
```diff
@@ -121,17 +130,23 @@ def notify(recipients,log_type):
...
-	if log_type=="OUT":
-		send_notification(checkout_subject, checkout_message, user_id)
-		push_notification(employee_id, Notification_title, "Please checkout in the next five minutes.")
+		user_id = recipient[0]
+		employee_id = recipient[1]
+		print(recipient)
+		if log_type=="IN":
+			send_notification(checkin_subject, checkin_message, user_id)
+			#arrive late button is true only if the employee is assigned with shift to head office.
+			if recipient[2] == "Head Office-Head Office-Morning|08:00:0-17:00:0|9 hours":
```
Ehtesham599: Avoid hard coding here. Try using a different approach to set the default shift for head office and fetch that shift from where you set it.
Changing the name of the shift in future will cause problems.

Ehtesham599 (changes requested):

one_fm/api/tasks.py
SalShai: done

one_fm/api/tasks.py
SalShai: Used User Role instead.

Ehtesham599 (approved):

PR #667, opened 2021-11-11
#180244038 ot operation payroll

Ehtesham599:
## Feature description
A function that creates an Additional Salary document with an "Overtime Allowance" component for an employee when an Attendance of roster type "Over-Time" is submitted.

## Solution description
Added a field roster_type in Shift Assignment and the Attendance doctypes. 

Once the Attendance document is submitted, the function checks is the roster_type of the attendance is "Over-Time".
Only if the roster_type is "Over-Time", it proceeds to obtain the shift and project details where the employee had worked in.
If the project has an overtime rate set, it computes the overtime allowance amount by multiplying the overtime rate by the number of hours worked in the shift.
Else if there is no overtime allowance set for the project, it follows the labor law to calculate the overtime allowance by first fetching the employee's basic salary, computing his/her hourly wage based on the shift the employee is assigned to and multiplying the hourly wage by number of worked hours in the shift and 1.5.
`basic hourly wage * number of worked hours * 1.5`

Once the amount is calculated, it proceeds to create an Additional Salary document with the Overtime Allowance component for that particular employee.

## Areas affected and ensured
Updated the feature of setting the shift details in the Attendance by including the roster_tye field in the same function and changed the controller type of this function call in hooks from `on_submit` to `on_save`.
Nothing seems to be affected and is ensured.

Shift details will be set upon saving the attendance document instead of submitting it.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
Approved👍🏼

PR #677, opened 2021-11-15
Pushing Body Parameter according to the OS of the device.

SalShai:
## Feature description
Pushing Body Parameter according to the OS of the device. Two device OS: Android and iOS require different parameters for the body. 

## Solution description
The Response is sent using a Rest API. The device OS is stored in Employee doctype along with the device Token. The Body is designed based on the device OS.

Screen Shot 2021-11-14 at 2 46 29 PM

## Areas affected and ensured
- Final Reminder 
- Push Notification

No

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/api.py
+++ b/one_fm/api/api.py
@@ -96,15 +96,18 @@ def rename_post(posts):
         except Exception as e:
             print(frappe.get_traceback())
 
-"""
-This function stores FCM Token  and Device OS in employee doctype 
-that is fetched from device/app end when user logs in.
-Params: Employee ID (Single employee ID), FCM Token and Device OS comes from the client side.
-It returns true or false based on the execution.
-eg:bench execute --kwargs "{'employee_id':'HR-EMP-00002','fcm_token':'cKg2nMfqwUpDinaWmUmVXp:APA91bHYAT2UxhesFtFHgweK8umoXm2LR9xyvi1L1OoefDRb3f1etOQbR3MZRFwL7Vwgc6hHrXO7urQWg3iy5SwTh_iE0lQP0Ny3O2nkcK_5Zo_UCQPMbZORF2Y5beejJbiHcQXA_x1C', 'device_os':'IOS'}" one_fm.api.api.store_fcm_token 
-"""
 @frappe.whitelist()
 def store_fcm_token(employee_id ,fcm_token,device_os):
+    """
+    This function stores FCM Token  and Device OS in employee doctype 
+    that is fetched from device/app end when user logs in.
+    
+    Params: Employee ID (Single employee ID), FCM Token and Device OS comes from the client side.
+    
+    It returns true or false based on the execution.
+    
+    eg:bench execute --kwargs "{'employee_id':'HR-EMP-00002','fcm_token':'f0_1sEWK7kksiegwCZ7dUm:APA91bHCFXcmXdI7AnI_37dbfjTz5uKf46_kvwIXgmtSoxGCbApo4zFETfbaWaEj8FpKXzJlwUS6CTTCfSX8WStuiFx1oPR4gtH3I46-jNbQkbhfvXI_lR3mKDl6e2ek2nB_5OFTfH9c', 'device_os':'ios'}" one_fm.api.api.store_fcm_token 
+    """
     Employee = frappe.get_doc("Employee",{"name":employee_id})
     try:
         if Employee:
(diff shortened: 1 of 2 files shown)


one_fm/api/api.py
```diff
@@ -96,15 +96,18 @@ def rename_post(posts):
...
 def store_fcm_token(employee_id ,fcm_token,device_os):
+    """
+    This function stores FCM Token  and Device OS in employee doctype 
+    that is fetched from device/app end when user logs in.
+    
+    Params: Employee ID (Single employee ID), FCM Token and Device OS comes from the client side.
+    
+    It returns true or false based on the execution.
+    
+    eg:bench execute --kwargs "{'employee_id':'HR-EMP-00002','fcm_token':'f0_1sEWK7kksiegwCZ7dUm:APA91bHCFXcmXdI7AnI_37dbfjTz5uKf46_kvwIXgmtSoxGCbApo4zFETfbaWaEj8FpKXzJlwUS6CTTCfSX8WStuiFx1oPR4gtH3I46-jNbQkbhfvXI_lR3mKDl6e2ek2nB_5OFTfH9c', 'device_os':'ios'}" one_fm.api.api.store_fcm_token
```
Ehtesham599: There's no need to provide a bench execute guide here.

one_fm/api/api.py
```diff
@@ -152,41 +158,62 @@ def push_notification(employee_id, title, body, checkin, arriveLate ,checkout ):
...
+    Params: employee_id e.g. HR_EMP_00001, , title:"Title of your message", body:"Body of your message"
+    test Execution: bench execute --kwargs "{'employee_id':'HR-EMP-00001','title':'Hello','body':'Testing','checkin':'True','arriveLate':'True','checkout':'False'}" one_fm.api.api.push_notification_rest_api
+
     serverToken is fetched from firebase -> project settings -> Cloud Messaging -> Project credentials
-    Device Token is store in employee doctype using 'store_fcm_token' on device end.
+    Device Token and Device OS is store in employee doctype using 'store_fcm_token' on device end.
     """
     serverToken = frappe.get_value("Firebase Cloud Message",filters=None, fieldname=['server_token'])
-    token = frappe.get_all("Employee", {"name": employee_id}, "fcm_token")
+    token = frappe.get_list("Employee", {"name": employee_id}, "fcm_token, device_os")
```
Ehtesham599: Use `frappe.db.get_value` instead.

one_fm/api/api.py
```diff
@@ -118,15 +121,18 @@ def store_fcm_token(employee_id ,fcm_token,device_os):
     except Exception as e:
         print(frappe.get_traceback())
 
-"""
+
+@frappe.whitelist()
+def push_notification(employee_id, title, body, checkin, arriveLate ,checkout):
```
Ehtesham599: This function is specific to check-in/arrive late. Follow appropriate naming for the method.

one_fm/api/api.py
```diff
@@ -152,41 +158,62 @@ def push_notification(employee_id, title, body, checkin, arriveLate ,checkout ):
...
     return response
 
-# This function is used to send notification through Firebase CLoud Message. 
-# It is a rest API that sends request to "https://fcm.googleapis.com/fcm/send"
-# Params: employee_id e.g. HR_EMP_00001, , title:"Title of your message", body:"Body of your message"
-# bench execute --kwargs "{'employee_id':'HR-EMP-00001','title':'Hello','body':'Testing'}" one_fm.api.api.push_notification_rest_api
 @frappe.whitelist()
-def push_notification_rest_api(employee_id, title, body):
-    """
+def push_notification_rest_api(employee_id, title, body, checkin, arriveLate ,checkout ):
```
Ehtesham599: Same here. Follow appropriate naming for the method. It can be misleading to other developers trying to use the same method to send push notifications.

one_fm/api/api.py
```diff
@@ -152,41 +158,62 @@ def push_notification(employee_id, title, body, checkin, arriveLate ,checkout ):
...
+    """ 
+    This function is used to send notification through Firebase CLoud Message. 
+    It is a rest API that sends request to "https://fcm.googleapis.com/fcm/send"
+    
+    Params: employee_id e.g. HR_EMP_00001, , title:"Title of your message", body:"Body of your message"
+    test Execution: bench execute --kwargs "{'employee_id':'HR-EMP-00001','title':'Hello','body':'Testing','checkin':'True','arriveLate':'True','checkout':'False'}" one_fm.api.api.push_notification_rest_api
```
Ehtesham599: Same here. No need to provide execution commands for any function in-line documentation.

Ehtesham599 (changes requested):

one_fm/api/api.py
SalShai: @Ehtesham599 Done!

Ehtesham599 (approved):

PR #685, opened 2021-11-17
Returning the Retries Left if face recognition Fails.

SalShai:
## Feature description
This is an API to accept penalties. To Accept a Penalty, one needs to pass the face recognition test. Image file in base64 format is passed through the face recognition test. And, the employee is given 3 tries. If face recognition is true, the penalty gets excepted. If Face recognition fails even after 3 tries, the image is sent to the legal manager for investigation. 

## Solution description
Send response using Response API. 
If Success = True, Status would be 200.
If Face recognition fails, return back the Retries left in data_obj.

On success:
Screen Shot 2021-11-16 at 7 44 32 PM

On Fail:
Screen Shot 2021-11-16 at 7 50 56 PM

## Areas affected and ensured
Accept Penalty API Response structure is Changed.

yes, Response for Accept_Penalty.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/mobile/legal.py
+++ b/one_fm/api/mobile/legal.py
@@ -176,9 +176,9 @@ def accept_penalty(file, retries, docname):
 
 			frappe.db.commit()
 
-			return response("Face Recognition Successfull.", True ,{},200)
+			return response("Face Recognition Successfull.", {}, True ,200)
 		else:
-			return response("Face Recognition Failed. You have "+str(retries_left)+" retries left.", False ,{"retries_left":retries_left},401)
+			return response("Face Recognition Failed. You have "+str(retries_left)+" retries left." ,{"retries_left":retries_left},False,401)
 			penalty.db_set("retries", retries_left)
 			frappe.throw(_("Face could not be recognized. You have {0} retries left.").format(frappe.bold(retries_left)), title='Validation Error')
 
@@ -203,14 +203,9 @@ def reject_penalty(rejection_reason, docname):
 			penalty.workflow_state = "Penalty Rejected"
 			penalty.save(ignore_permissions=True)
 			frappe.db.commit()
-			return {
-					'message': 'success'
-				}
+			return response("Penalty Rejected", {}, True, 200)
 		else:
-			return {
-					'message': f'No penalty {docname} found'
-				}
+			return response("No penalty {docname} found." , {}, False, 401)
 	except Exception as exc:
-		print(frappe.get_traceback())
 		frappe.log_error(frappe.get_traceback())
-		return frappe.utils.response.report_error(exc)
+		return response(exc,False,[], 500)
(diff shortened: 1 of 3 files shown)


one_fm/api/mobile/legal.py
```diff
@@ -172,17 +178,15 @@ def accept_penalty(file, retries, docname):
 
 			frappe.db.commit()
 
-			return {
-				'message': 'success'
-			}
+			return response("Face Recognition Successfull.", True ,{},200)
 		else:
+			return response("Face Recognition Failed. You have "+str(retries_left)+" retries left.", False ,{"retries_left":retries_left},401)
```
Amna-Alshawa: Try to set the response args in this way:
**response( `message` , `data` ,  `success` , `status code` )** 
to unite all api responses.

one_fm/utils.py
```diff
@@ -2036,4 +2036,12 @@ def create_additional_salary(employee, amount):
 	additional_salary.overwrite_salary_structure_amount = 1
 	additional_salary.notes = "Overtime Earning"
 	additional_salary.insert()
-	additional_salary.submit()
\ No newline at end of file
+	additional_salary.submit()
+
+def response(message, success, data, status_code):
```
Amna-Alshawa: Make sure you update the parameters in the header definition of the `response`  method.

one_fm/api/mobile/legal.py
```diff
@@ -148,13 +154,13 @@ def accept_penalty(file, retries, docname):
...
 		penalty = frappe.get_doc("Penalty", docname)
 		image = upload_image(file, OUTPUT_IMAGE_PATH)
-		if recognize_face(file) or retries_left == 0:
+		if recognize_face(image) or retries_left == 0:
 			if retries_left == 0:
 				penalty.verified = 0
 				send_email_to_legal(penalty)
 			else:
 				penalty.verified = 1		
-			penalty.workflow_state = "Penalty Accepted"
```
Amna-Alshawa, line 157: @SalShai ,you are accepting penalties even before the three tries!!

one_fm/api/mobile/legal.py
```diff
@@ -134,6 +135,11 @@ def get_penalty_details(penalty_name):
 @frappe.whitelist()
 def accept_penalty(file, retries, docname):
 	"""
+	This is an API to accept penalty. To Accept Penalty, one needs to pass the face recognition test.
+	Image file in base64 format is passed through face regonition test. And, employee is given 3 tries.
+	If face recognition is true, the penalty gets excepted.
```
Amna-Alshawa: Do you mean the penalty gets `accepted`, so `excepted` is a typo issue?

Amna-Alshawa (changes requested):
@SalShai Also, remove all print in `one_fm/legal/doctype/penalty/penalty.py` file

one_fm/api/mobile/legal.py
SalShai: how does that make any difference ?

Amna-onefm: no difference but organizing and unifying API  responses.

Amna-Alshawa (approved):
@SalShai Greate

PR #686, opened 2021-11-17
#180327885

SalShai:
## Feature description
Fetching Leave Approver and Sending Leave Approval Notification in HTML format. This notification would consist of 2 actions: Accept or Reject.

## Solution description
Fetch the correct leave approver: either Report_to or Leave Approver. 
    But, if both don't exist, the Operation manager is the Leave Approver.
Once the form is submitted, an Approval Notification is sent through email. 

Screen Shot 2021-11-17 at 12 44 40 PM

## Areas affected and ensured
Leave Application Notification:
- if Leave Type is Sick Leave, then text formate email is sent without Buttons.
- if Not, the doc status is open, and the HTML format is submitted, with Action Buttons.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/mobile/Leave_application.py
+++ b/one_fm/api/mobile/Leave_application.py
@@ -167,7 +167,7 @@ def notify_leave_approver(doc):
     This function is to notify the leave approver and request his action. 
     The Message sent through mail consist of 2 action: Approve and Reject.(It is sent only when the not sick leave.)
 
-    Param: doc -> Leave Application DocName (which needs approval)
+    Param: doc -> Leave Application Doc (which needs approval)
 
     It's a action that takes place on update of Leave Application.
     """
--- a/one_fm/api/mobile/Leave_application.py
+++ b/one_fm/api/mobile/Leave_application.py
@@ -167,6 +167,8 @@ def notify_leave_approver(doc):
     This function is to notify the leave approver and request his action. 
     The Message sent through mail consist of 2 action: Approve and Reject.(It is sent only when the not sick leave.)
 
+    Param: doc -> Leave Application DocName (which needs approval)
+
     It's a action that takes place on update of Leave Application.
     """
     #If Leave Approver Exist
--- a/one_fm/api/mobile/Leave_application.py
+++ b/one_fm/api/mobile/Leave_application.py
@@ -3,7 +3,6 @@ from frappe import _
 from erpnext.hr.doctype.leave_application.leave_application import get_leave_balance_on, get_leave_allocation_records, get_leave_details
 from datetime import date
 import datetime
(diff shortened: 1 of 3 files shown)


one_fm/api/mobile/Leave_application.py
```diff
@@ -134,4 +133,60 @@ def response(message, data, status_code):
...
+    """
+    approver = None
+    #check if employee has leave approver or report to assigned in the employee doctype
+    leave_approver, report_to = frappe.db.get_value("Employee", employee, ["leave_approver", "reports_to"])
+    if not leave_approver:
+        if report_to:
+            approver = frappe.db.get_value('Employee', {'name': report_to}, ['user_id'])
+        else:
+            #if not, return the 'Operational Manager' as the leave approver. But, check if employee himself is not a leave manager.
+            operation_manager = frappe.db.get_value('Employee', {'Designation': "Operation Manager"}, ['name','user_id'])
```
Amna-Alshawa, line 158: I guess filtering by designation is not a good practice, what if multiple employees have the same designation then who will approve the leave?

Amna-Alshawa, line 158: I guess filtering by designation is not a good practice. What if multiple employees have the same designation, then who will approve the leave?

one_fm/api/mobile/Leave_application.py
```diff
@@ -134,4 +133,60 @@ def response(message, data, status_code):
...
+            if operation_manager[0]!= employee:
+                approver = operation_manager[1]
+    else:
+        approver = leave_approver
+    return approver
+
+def notify_leave_approver(doc):
```
Amna-Alshawa, line 165: Define `doc` in the comment that you are passing the `leave` object.

Amna-Alshawa (changes requested):

one_fm/api/mobile/Leave_application.py
SalShai: Though there could be multiple operation managers, the approval could go to any one of them. Since Both are eligible to give approval for leave.

SalShai: Moreover, this solution is not a permanent one and has high chances of modification once planned well. In the Future, It probably would go to the Supervisors or everyone would be assigned with a "Report_to".

Amna-Alshawa: @SalShai I do understand that this is not a permanent solution, but is there a chance where you can fetch the operation manager from the shift or employee doctype instead of notifying all operations managers?
As the notification won't go for `any` it will go for `all` of them!

Amna-Alshawa (approved):
@SalShai , greate work

PR #692, opened 2021-11-18
Fix: workflow default email alert

Amna-Alshawa:
## Feature description
Disable default email alert for workflow

Disable default email alert for workflow

## Areas affected and ensured
No

No

  - [x] Chrome
  - [ ] Safari


SalShai (approved):
@Amna-Alshawa, Done.

PR #696, opened 2021-11-22
Add Firebase_admin Initialization inside a function

SalShai:
## Feature description
Create a function to initialize Firebase-Admin

## Solution description
The firebase admin was initialized independently, which was causing an issue. So, we created a new function that is been called only when necessary.

## Areas affected and ensured
Push Notifications for Checkin
initialize_firebase

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/api.py
+++ b/one_fm/api/api.py
@@ -10,8 +10,14 @@ import json
 from frappe.desk.page.user_profile.user_profile import get_energy_points_heatmap_data, get_user_rank
 from frappe.social.doctype.energy_point_log.energy_point_log import get_energy_points, get_user_energy_and_review_points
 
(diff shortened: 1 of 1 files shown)


one_fm/api/api.py
```diff
@@ -10,8 +10,14 @@
...
+def initialize_firebase():
+    """
+        Initialize Firebase-Admin
+    """
+    #fetch credential file to initilize Firebase SDK
+    cred = credentials.Certificate(frappe.utils.cstr(frappe.local.site)+"/private/files/one-fm-70641-firebase-adminsdk-nuf6h-667458c1a5.json")
+    firebase_admin.initialize_app(cred)
 
 @frappe.whitelist()
 def _one_fm():
```
Amna-Alshawa, line 23: when does this method been called?
why do you name it like this?

Amna-Alshawa (approved):
Approved

PR #700, opened 2021-11-24
[Feat] Proof Document Required for Apply Leave

pjamsheer:
## Feature description
Employee needs to attach an attachment when applying for a leave, so employee can provide proof for the leave request

## Solution description
- Customize Leave Type to add a check for Proof document required
- Customize the Leave Application DocType to add an attach field based on the Leave Type configuration

Screenshot 2021-11-25 at 1 28 26 AM
Screenshot 2021-11-25 at 1 11 56 AM
Screenshot 2021-11-25 at 1 14 40 AM
Screenshot 2021-11-25 at 1 16 15 AM

## Areas affected and ensured
Updated `hooks.py` and `hr_utils.py`

No

  - [x] Chrome
  - [x] Safari


Amna-Alshawa (approved):
@pjamsheer 👍

PR #714, opened 2021-12-12
#180458083-backend

Ehtesham599:
Refactored backend for civil ID OCR using vision API.


Ehtesham599:
@SalShai Check out the changes I have made and added an extra argument in `get_civil_id_text` method.

SalShai (approved):

PR #717, opened 2021-12-13
[Fix] Career History - Doctype and Portal Page

pjamsheer:
## Feature description
- Magic Link - Having `get_magic_link`, `get_encrypted_magic_link`, `create_magic_link`, `authorize_magic_link` and `send_magic_link` methods.!!!
- Career History - Doctype - Updated and Optimised
- Career History Portal Page Redesigned
- Magic Link can be send from Job Applicant for Career History

- Changed calculations of Experience and Promotion in Career History

## Solution description
- Removed All Useless codes from the Old Career History Doctype and Improved for calculation purpose
Screenshot 2021-12-13 at 3 59 02 AM
Screenshot 2021-12-13 at 3 59 14 AM

- Encrypted Magic Link is Developed
- Career History Page completely Redesigned
https://user-images.githubusercontent.com/20554466/145738668-c4eec301-1c54-4185-a9d5-efad5154ea51.mov

## Areas affected and ensured
- Career History - Implementation changed - Removed Unused Child tables and Updated the company table in the Career History
- Updated Career History Portal Page

No.

  - [x] Chrome
  - [x] Safari


one_fm/one_fm/doctype/magic_link/magic_link.py
```diff
@@ -0,0 +1,107 @@
...
+		Method is used to check the magic ling exist or expired
+		args:
+			encrypted_magic_link: encrypted value of Magic Link ID
+			doctype: Reference Doctype name, from which the magic link created
+			link_for: The doctype for which the magic link is created (example: Career History)
+
+		return Magic Link ID (example: ML.000001)
+	'''
+	decrypted_magic_link = False
+	try:
```
Ehtesham599, line 72: Why a try catch block for a `decrypted_magic_link` True/False condition?
You can have a condition check for decrypted magic link or not inside the try block and in the Exception block you can throw/log the exception message instead.

Ehtesham599 (changes requested):

one_fm/one_fm/doctype/magic_link/magic_link.py
pjamsheer: In decrypt method it through error if the encrypted link is edited, I am not able to catch that and show the error if I didn't use the try. Tried conditional checking and all and ended up with this.

Ehtesham599 (approved):
LGTM.

PR #723, opened 2021-12-13
[Feat] Create Leave Policy Assignment, while existing one expires

pjamsheer:
## Feature description
- Create Leave Policy Assignment, while existing one expires

## Solution description
 - On Daily Scheduler - Get all Leave Policy Assignments, those expires today and then duplicate it with changing the effective from and effective to dates

No

  - [x] Chrome
  - [x] Safari


Ehtesham599 (approved):

Ehtesham599:
@pjamsheer Resolve conflicts.

PR #730, opened 2021-12-15
Make changes in UI

SalShai:
## Feature description
Create UI for Applicant docs using bootstrap.

## Solution description
- Use Bootstrap for UI
- Call some function on Window Load.
- Change input Type for "Gender" and "Passport Type"

Screen Shot 2021-12-15 at 3 51 04 PM

## Areas affected and ensured
UI

- UI has a change based on Nationality. I.e. if the non-Kuwaiti, the form has an extra field for the Sponsor's Name.
- Filling of Nationality select options happens on window load.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/public/assets/css/applicant_docs.css
+++ b/one_fm/public/assets/css/applicant_docs.css
@@ -1,9 +1,8 @@
 .main-container{
-    width: 68vw;
+    width: 63vw;
     height: auto;
-    padding: 11vh;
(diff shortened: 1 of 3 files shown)


pjamsheer:
@SalShai Grate, is it on the top of latest master-13. If it is not, then do  a rebase against master-13 and push

pjamsheer (approved):

PR #737, opened 2021-12-21
#180610682: Electronic Signature Declaration

SalShai:
## Feature description
As an Employee, I want to have my signature in an electronic format, so that I can use it to complete all company processes in ERPNext that require my signature

## Solution description
- Create Doctype named "Electronic Signature Declaration" from Onboarding. Link it back to Onboarding Doctype.
- Update Status of the signature on Onboarding Doctype. Proceed with work contract only when applicant signed.
- Create Print Format for the same.

Screen Shot 2021-12-21 at 10 31 13 AM
Screen Shot 2021-12-21 at 10 31 40 AM
Screen Shot 2021-12-21 at 10 32 29 AM

## Areas affected and ensured
- New Doctype to store Employee Signature.
- Follow up with work contract only after the doc is signed.
- Fetch the value to employee doc.

Yes.
- New Doctype to store Employee Signature.
- Follow up with work contract only after the doc is signed.
- Fetch the value to employee doc.
- 
  - [x] Chrome
  - [x] Safari


Ehtesham599 (approved):
LGTM!

PR #741, opened 2021-12-21
#180610685

Ehtesham599:
## Feature description
As a Company, I want to have a Duty Commencement DocType, so that we can facilitate the switch between Onboarding to Employee.

## Solution description
When the Duty Commencement workflow is set as "Applicant Signed and uploaded" and an employee record is created for the candidate from onboarding and if the date of joining is today, a Shift Assignment record is created in order to allow the employee to checkin to the system. If the current time is past the shift start time,  the employee is automatically checked in into the system.
Note: When the employee is created from onboarding, this employee is set in Duty commencement.

## Output screenshots
Duty Commencement:
Screen Shot 2021-12-21 at 1 16 08 PM
Screen Shot 2021-12-21 at 1 16 30 PM

Shift Assignment:
Screen Shot 2021-12-21 at 1 17 03 PM

Employee Checkin:
Screen Shot 2021-12-21 at 1 17 33 PM

## Areas affected and ensured
Nothing affected.

No.

  - [x] Chrome
  - [x] Safari


pjamsheer:
@Ehtesham599 There is a conflict

Ehtesham599:
@pjamsheer Check now

pjamsheer (approved):

PR #748, opened 2021-12-22
#180692998 day off job offer

Ehtesham599:
## Feature description
As a Roster Operator, I want to specify the number of days off for each employee in the Job Offer and Onboarding Doctype, so that data can be used to calculate the OT once the candidate starts working.

## Solution description
Added Day Off Category and Number Of Days Off to Job Offer and Onboard Employee and added validation on the same.
Day Off Category is of two types: 
- Weekly
- Monthly

## Areas affected and ensured
Nothing affected.

No.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
Approved!

PR #749, opened 2021-12-23
Indicate roster day off OT

Ehtesham599:
## Feature description
As a Roster Operator, I want to indicate when an employee is working on his Day off, so that the employee receives an OT pay for that day

## Solution description
Added a checkbox in Employee Schedule to set if employee is working on a day off or not.
Similar checkbox added in the Schedule/Change Post dialog to make a day off OT entry.
Note: Day off OT can only be scheduled from the OT Roster tab.

## Output screenshots
Schedule/Change Post dialog OT Roster tab:
Screen Shot 2021-12-23 at 9 49 54 AM

Day Off OT indicator in OT roster tab:
Screen Shot 2021-12-23 at 9 50 39 AM

## Areas affected and ensured
Nothing affected and ensured.

No.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
Approved! But, could you please write the summary for the function??

PR #750, opened 2021-12-23
Salary cash export

Ehtesham599:
## Feature description
As a Payroll Officer, I want to process cash payroll, so that it doesn't get paid automatically with the bank transfer process

## Solution description
Added salary mode to Payroll Employee Detail and fetched this salary mode from employee record.
Implemented a method to get list of employees from payroll entry that have salary mode as "Cash" and export their details into an excel file.
Added button in Payroll Entry frontend to download the exported file.

## Output screenshots
Download option:
Screen Shot 2021-12-23 at 3 14 20 PM

Exported file:
Screen Shot 2021-12-23 at 3 15 03 PM

## Areas affected and ensured
Nothing affected.

No.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
Done!

PR #753, opened 2021-12-27
Fix: Download Issue

Ehtesham599:
- Changed URL from "public/file/payroll-entry" to "files/payroll-entry".
- Removed backend method to fetch file URL.
- Directly used window,open() instead of making a backend call to fetch file URL.


SalShai (approved):
Done.

PR #773, opened 2022-01-03
fix: RFM item_code display

mymi14s:
This is a fix that displays item_code(Item) in 'Request for Material Item' table.

- Screenshot
image


pjamsheer:
@mymi14s This is not the required fix.
Product owner needs to show the item_code field only after the Approval of Request for Material

mymi14s:
> @mymi14s This is not the required fix. Product owner needs to show the item_code field only after the Approval of Request for Material

I will rework and revert.

PR #780, opened 2022-01-04
[Fix] Add Quotation Attachments to the PO

pjamsheer:
## Feature description
 - Add Quotation Attachments to the PO
 - Fixes in RFM

## Solution description
 - Get attachments from the Quotation and Update those attachments to the PO in after insert method
 - Created a method `set_attachments_to_doctype` for fetch attachments from source and set to the target doc

No

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -154,7 +154,9 @@ doc_events = {
 	"Purchase Order": {
 		"on_submit": "one_fm.purchase.doctype.request_for_material.request_for_material.update_completed_purchase_qty",
 		"on_cancel": "one_fm.purchase.doctype.request_for_material.request_for_material.update_completed_purchase_qty",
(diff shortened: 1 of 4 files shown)


one_fm/hooks.py
```diff
@@ -154,7 +154,9 @@
 	"Purchase Order": {
 		"on_submit": "one_fm.purchase.doctype.request_for_material.request_for_material.update_completed_purchase_qty",
 		"on_cancel": "one_fm.purchase.doctype.request_for_material.request_for_material.update_completed_purchase_qty",
-		"on_update_after_submit": "one_fm.purchase.utils.check_for_signature_for_purchase_order"
+		"on_update_after_submit": "one_fm.purchase.utils.check_for_signature_for_purchase_order",
```
SalShai, line 157: Shouldn't the methods be inside a square bracket?

SalShai (approved):
Done👍🏼

PR #786, opened 2022-01-05
Feat: Report of Approved Purchase Document

SalShai:
## Feature description
Report of the documents approved in the Purchase Process.

## Solution description
Create a report using Version Doctype to fetch all the doctypes and the doc link, if it was approved by the user.

Screen Shot 2022-01-06 at 8 45 45 AM

## Areas affected and ensured
- New Report Created.
- Enable "Track The Changes" in the 'Request for Purchase' Doctype.
-
-Changes tracked in 'Request for Purchase' Doctype.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/one_fm/report/purchase_process/purchase_process.js
+++ b/one_fm/one_fm/report/purchase_process/purchase_process.js
@@ -1,0 +1,9 @@
+// Copyright (c) 2016, omar jaber and contributors
+// For license information, please see license.txt
+/* eslint-disable */
+
+frappe.query_reports["Purchase Process"] = {
+	"filters": [
+		
+	]
+};
--- a/one_fm/one_fm/report/purchase_process/purchase_process.json
+++ b/one_fm/one_fm/report/purchase_process/purchase_process.json
@@ -1,0 +1,30 @@
+{
+ "add_total_row": 0,
+ "columns": [],
+ "creation": "2022-01-05 11:24:16.707795",
+ "disable_prepared_report": 0,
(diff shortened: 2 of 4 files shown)


mymi14s:
Please add the workflow_state (original_value, new_value) to the report columns.
This will report the old and new value of the approval, Thanks.

image

mymi14s (changes requested):
Also return the report with values sorted by approval date DESC

SalShai:
@mymi14s Could you elaborate more on why do we need old values? While we could just be displaying the current status.

SalShai:
Sure, let me sort the approval date by DESC.

mymi14s:
> @mymi14s Could you elaborate more on why do we need old values? While we could just be displaying the current status.

So that the report can show what action was taken from x to y.

SalShai:
The action is always from Draft to Approved😅

SalShai:
> > @mymi14s Could you elaborate more on why do we need old values? While we could just be displaying the current status.
> 
> So that the report can show what action was taken from x to y.

@mymi14s , The action is always from Draft to Approved😅

SalShai:
> > > @mymi14s Could you elaborate more on why do we need old values? While we could just be displaying the current status.
> > 
> > 
> > So that the report can show what action was taken from x to y.
> 
> @mymi14s , The action is always from Draft to Approved😅

@taqi-onefm , Could you please Clarify the doubt??

mymi14s (approved):
Good to go

PR #794, opened 2022-01-10
chore: Auto download facial recognition file before installation

mymi14s:
## Auto download facial recognition file before installation
One FM app require facial recognition for processing images, a .dat file 'shape_predictor_68_face_landmarks.dat' for the processing images. Fresh installation of ONE-FM app fails because the file was not found.

This PR checks if the file exists before installation/app update, if exists, skips else it downloads and save the file in the private directory.

The file can be found in the following url:  [https://github.com/italojs/facial-landmarks-recognition/blob/master/shape_predictor_68_face_landmarks.dat](url)

An installation drive was executed and result can be seen in screenshot below.
image

### DIFF
--- a/one_fm/__init__.py
+++ b/one_fm/__init__.py
@@ -14,7 +14,7 @@ from one_fm.api.mobile.Leave_application import notify_leave_approver
 from one_fm.hiring.utils import grant_leave_alloc_for_employee
 
 
-__version__ = '0.0.1'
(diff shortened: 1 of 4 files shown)


mymi14s:
Text 'saving to Landmark facial recogintion to' has been corrected to 'saving Landmark facial recognition to''

Ehtesham599 (approved):
LGTM.

PR #797, opened 2022-01-11
[Fix] RFM - Items field property - based on status

pjamsheer:
## Feature description
- RFM - Items field property - based on status

## Solution description
- Updated client side code form property setter
- Optimised the property setter code

## Areas affected and ensured
 - `request_for_material.js`

Yes, Read only property of RFM Item table changes according to the status and docstatus

  - [x] Chrome


SalShai (approved):
Approved!

PR #798, opened 2022-01-11
fix: payroll entry, issue with 'Get Employee' failed

mymi14s:
## Feature description
This is a fix for payroll entry 'Get Employee' button, the custom button retrieve employee(s), set in child table. This fails because a custom validation could not skip employees with no bank account nor salary mode.

## Solution description
The fix check for employee(s) with no salary mode or bank account, sends the information via email to HR Manager/User, display same on the screen, create issue ticket and allows the process to continue.

## Screenshots
image

image

### DIFF
--- a/one_fm/api/doc_methods/payroll_entry.py
+++ b/one_fm/api/doc_methods/payroll_entry.py
@@ -110,7 +110,7 @@ def set_bank_details(employee_details):
 		)
 		frappe.msgprint(_(message))
 		# send and log missing payment detail
-		# frappe.enqueue(method=email_missing_payment_information, queue='short', timeout=300, **{'message':message})
+		frappe.enqueue(method=email_missing_payment_information, queue='short', timeout=300, **{'message':message})
 	return employee_details
 
 def get_count_employee_attendance(self, employee):
--- a/one_fm/api/doc_methods/payroll_entry.py
+++ b/one_fm/api/doc_methods/payroll_entry.py
@@ -91,11 +91,10 @@ def set_bank_details(employee_details):
 			else:
 				iban, bank, bank_account_no = None, None, None
 
-			if(salary_mode=='Bank' and not bank):
+			if(salary_mode.lower()=='bank' and bank is None):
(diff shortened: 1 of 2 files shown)


one_fm/api/doc_methods/payroll_entry.py
```diff
@@ -79,25 +79,45 @@ def set_bank_details(employee_details):
...
+			if not salary_mode:
+				employee_missing_detail.append({'employee':employee, 'salary_mode':'', 'issue':'No salary mode'})
+			employee.salary_mode = salary_mode
+			employee.iban_number = iban or bank_account_no
+			bank_code = frappe.db.get_value("Bank", {'name': bank}, ["bank_code"])
+			employee.bank_code = bank_code
+		except Exception as e:
+			frappe.log_error(str(e), 'Payroll Entry')
+			frappe.throw(str(e))
+	# print(employee_missing_detail)
```
Ehtesham599, line 112: You can remove this commented print statement

Ehtesham599 (commented):
@mymi14s Will this allow the payroll entry to be submitted without getting the details of all the employees?

one_fm/api/doc_methods/payroll_entry.py
mymi14s: Yes

mymi14s: I meant, it will fetch the employees normally,  but inform HR about missing salary mode. Before it generates errors in background and fails but now it will process and report as well, what do you think?

Ehtesham599:
@mymi14s Instead of removing the employee from the payroll, don't allow the payroll entry to be submitted if all the employees don't have the required data.
I think that should complete the fix.

mymi14s:
> @mymi14s Instead of removing the employee from the payroll, don't allow the payroll entry to be submitted if all the employees don't have the required data. I think that should complete the fix.

That is a good suggestion, I will work on it.

mymi14s:
Issue handle in https://github.com/ONE-F-M/One-FM/pull/818

PR #802, opened 2022-01-12
chore: create project-based (one-time event) contracts.

mymi14s:
## Feature description
As a Finance Officer, I want to create project-based (one-time event) contracts, so that I can track them on the same system with my other periodic contracts.

This chore adds a custom field (Engagement Type) to **Contracts Docype** that can be used to specify if the contract is "One-of or Periodic". This helps to generated reports.

## Solution description
Engagement Type is a compulsory select field with two options:

- One-off
- Periodic
If One-off is selected, the "Is Auto Renewal" field is disabled and removed from the form else it remains in the form and enable.

image
image


Ehtesham599 (approved):
LGTM

PR #805, opened 2022-01-16
feat: Print Format for Contracts Sales Invoice Attendance (Absent/Present)

mymi14s:
## Feature description
As a Finance Officer, I want to have print format in Sales Invoice  for Contracts Attendance history attached with invoice (absent/present), so that I can invoice the client in the way they want

## Solution description
This PR is a print format design that appends employees attendance to sales invoice with a linked contracts. Unlike the traditional print format, this is generated from the backend and the content is render into the print page.

image


mymi14s (commented):
Some un-needed fixtures removed.

one_fm/jinja/print_format/methods.py
```diff
@@ -0,0 +1,131 @@
...
+        template, context = sic_attendance_absent_present(doc)
+        return frappe.render_template(
+            template, context
+        )
+
+pf = PrintFormat()
+
+
+# ATTENDANCE MAPS
+attendance_map = {
```
Ehtesham599, line 24: We already have abbreviations for attendance map in the roster something like below:

attendance_abbr_map = {
  'Present': 'P',
  'Absent': 'A',
  'Work From Home': 'WFH',
  'Half Day': 'HD',
  'On Leave': 'OL'
};

I recommend using that.

one_fm/utils.py
```diff
@@ -2303,4 +2303,16 @@ def generate_code():
 
 frappe.whitelist()
 def fetch_employee_signature(user_id):
-    return frappe.get_value("Employee", {"user_id":user_id},["employee_signature"])
\ No newline at end of file
+    return frappe.get_value("Employee", {"user_id":user_id},["employee_signature"])
+
+
+# date utilities
+class DateUtil:
```
Ehtesham599: To get first and last day of month, we can use frappe.utils.get_first_day(date) and  frappe.utils.get_last_day(date).
No need to create a class for this.

one_fm/jinja/print_format/methods.py
mymi14s: The PF have specific map.

Ehtesham599: We could go with anything tbh. Let's stick with what is already implemented in the application.
Just provide a legend in the print format for the attendance map.

pjamsheer (approved):

PR #818, opened 2022-01-31
chore: Clean up HR module

mymi14s:
## Feature description
This is a fix for payroll entry 'Get Employee' button, the custom button retrieve employee(s), set in child table. This fails because a custom validation could not skip employees with no bank account nor salary mode.

## Solution description
The fix check for employee(s) with no salary mode or bank account during Payroll Entry, log the information to 'Missing Payroll Information doctype, halt processes and display same on the screen, plus a button with link to the document with the missing detail the process to continue.

## Screenshots
image
image


taqi-onefm (approved):

taqi-onefm:
This is wonderful work @mymi14s !

PR #827, opened 2022-02-07
#181168691 ops chore

Ehtesham599:
## Feature description
Chore: Operations Adjustments 

## Solution description
- Set project type as mandatory in Project.
- Set POC mandatory in Project if project type is external.
- Set POC mandatory in Operations Site.


SalShai (approved):
Approved

PR #830, opened 2022-02-07
Sales Invoice from Contracts

Ehtesham599:
# Feature description
This feature supports generating invoices from contracts considering the invoice types configured in the contracts.
Supports:
- Single invoice for all items.
- Single invoice with separate item lines for each site.
- Separate invoice for each site.


mymi14s (approved):
Good job.

PR #833, opened 2022-02-08
Indemnity Allocation Calculation Report

SalShai:
## Feature description
- Create a new indemnity allocation to the active employee, if the record already doesn't exist.
- allocate new indemnity amount to the existing record.
- Create Report for indemnity Allocation

## Solution description
- If Indemnity Allocation for a given employee doesn't exist, create a new record. The indemnity amount of the employee would be based on his joining date till the current date.
- If the record exists, increment the amount as per day calculations.
- Report created to represent Indemnity Allocation for all the employees. 

https://user-images.githubusercontent.com/29017559/152968734-72679637-2e27-4138-a103-0f734c429164.mov

## Areas affected and ensured
- Indemnity allocation calculation formula has been changed.
- Few changes in the Report filter and columns.

- no

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/one_fm/hr_utils.py
+++ b/one_fm/one_fm/hr_utils.py
@@ -8,13 +8,14 @@ from frappe.model.document import Document
 
 def daily_indemnity_allocation_builder():
     query = """
-        select emp.name, emp.date_of_joining, emp.indemnity_policy
+        select emp.name, emp.date_of_joining
         from `tabEmployee` emp
         left join `tabIndemnity Allocation` ia on emp.name = ia.employee and ia.docstatus = 1 and emp.status = 'Active'
         where ia.employee is NULL
     """
     employee_list = frappe.db.sql(query, as_dict=True)
-    frappe.enqueue(indemnity_allocation_builder, timeout=600, employee_list=employee_list)
+    print(employee_list)
+    # frappe.enqueue(indemnity_allocation_builder, timeout=600, employee_list=employee_list)
 
 def indemnity_allocation_builder(employee_list):
     for employee in employee_list:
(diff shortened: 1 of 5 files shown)


mymi14s (approved):

one_fm/one_fm/hr_utils.py
```diff
@@ -21,38 +27,96 @@ def indemnity_allocation_builder(employee_list):
         create_indemnity_allocation(employee)
 
 def create_indemnity_allocation(employee):
-    indemnity_allcn = frappe.new_doc('Indemnity Allocation')
-    indemnity_allcn.employee = employee.name
-    indemnity_allcn.from_date = employee.date_of_joining
-    indemnity_allcn.new_indemnity_allocated = 30/365
-    indemnity_allcn.total_indemnity_allocated = 30/365
```
pjamsheer, line 28: @SalShai @mymi14s This is not Indemnity amount, This is the Indemnity factor, No need to calculate the amount instead of this !!

one_fm/one_fm/hr_utils.py
```diff
@@ -21,38 +27,96 @@ def indemnity_allocation_builder(employee_list):
...
-            (%s between from_date and to_date) and docstatus = 1 and status = 'Rejected'""",
-			(allocation.employee, getdate(nowdate())), as_dict=True)
-        if leave_appln and len(leave_appln) > 0:
-            allow_allocation = False
-
-        if allow_allocation:
-            # Set Daily Allocation
-            allocation.new_indemnity_allocated = 30/365
-            allocation.total_indemnity_allocated = allocation.total_indemnity_allocated+allocation.new_indemnity_allocated
-            allocation.save()
```
pjamsheer, line 55: @SalShai @mymi14s Asper the requirement, Indemnity will not be allocated for absent or Leave days !
Also Here we are calculating the factor of indemnity not the amount

mymi14s: @pjamsheer please check up with @SalShai as the story was accepted, if any differences, it can be hotfixed.

SalShai: @pjamsheer , They asked us to ignore the fact that the employee was absent. They are just considering the no. of days he is working for the company.

pjamsheer: Okay, Great.

one_fm/one_fm/hr_utils.py
SalShai: @pjamsheer According to the new requirement, Indemnity was calculated based on his no. of working years. if no. of working year is 5 or less, he gets 15 day's pay. above 5, he gets 30. And calculation involves taking per day payment(i.e. salary/26).

PR #836, opened 2022-02-09
#181199804: Fix: Check_exist checkin issue

SalShai:
## Feature description
Fix: Check_exist checkin issue

## Solution description
- Get log for the previous day and current day.  The previous day is taken into consideration for the Night shift.
- Get the last log type, so the next log type should be checkin.

NO.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/tasks.py
+++ b/one_fm/api/tasks.py
@@ -47,7 +47,7 @@ def send_checkin_hourly_reminder():
 				AND tSA.docstatus=1
 			""".format(date=cstr(date), shift_type=shift.name), as_list=1)
 			recipients = [recipient[0] for recipient in recipients if recipient[0]]
-			print(recipients)
+
 			subject = _("Hourly Reminder: Please checkin")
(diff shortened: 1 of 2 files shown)


one_fm/api/tasks.py
```diff
@@ -167,23 +167,6 @@ def notify_checkin_checkout_final_reminder(recipients,log_type):
 	elif log_type == "OUT":
 		send_notification(title, checkout_subject, checkout_message, notification_category, user_id_list)
 
-def insert_Contact():
```
pjamsheer, line 170: Is there any usage for this method

pjamsheer:
@SalShai You have inline comment and is this PR to `master-13` branch

one_fm/api/tasks.py
SalShai: No, This method isn't used anywhere. It was written by me to do some data entry.

pjamsheer (approved):

PR #837, opened 2022-02-15
#181199295 : Rearranging of the button.

SalShai:
## Feature description
Re-arrange Buttons in Job Applications.

## Solution description
- All the buttons are rearranged into one group. 
- Removing of creating Interview button by ERP since not being used.

Screen Shot 2022-02-15 at 12 07 24 PM

## Areas affected and ensured
Buttons are formed under one group.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/public/js/doctype_js/job_applicant.js
+++ b/one_fm/public/js/doctype_js/job_applicant.js
@@ -11,19 +11,19 @@ frappe.ui.form.on('Job Applicant', {
 	},
 	refresh(frm) {
 		// Changes the buttons for `PAM File Number` and `PAM Desigantion` once operator wants to changethe data of any
-		// if(frm.doc.pam_number_button == 0 || frm.is_new()){
-		// 	document.querySelectorAll("[data-fieldname='one_fm_change_pam_file_number']")[1].style.backgroundColor ="#3789ff";
-		// }if(frm.doc.pam_designation_button == 0 || frm.is_new()){
-		// 	document.querySelectorAll("[data-fieldname='one_fm_change_pam_designation']")[1].style.backgroundColor ="#3789ff";
-		// 	document.querySelectorAll("[data-fieldname='one_fm_change_pam_designation']")[1].style.marginLeft ='2em';
-		// }if(frm.doc.pam_number_button == 1 || !frm.is_new()){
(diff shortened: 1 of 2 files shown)


one_fm/public/js/doctype_js/job_applicant.js
```diff
@@ -37,67 +37,93 @@ frappe.ui.form.on('Job Applicant', {
...
+			frm.add_custom_button(__(''), function() {
+			},'Action').css({"padding": "0.01rem", "background-color":"gray"});
+
+			// view Interview button only when career history exist.
+			if(frm.doc.one_fm_job_applicant_score.find(i => i.reference_dt == "Interview Result")){
+				frm.add_custom_button(__('View Interview'), function() {
+					view_interview(frm);
+				},'Action');
+			}
+				frm.add_custom_button(__('Create An Interview'), function() {
```
Ehtesham599, line 78: Missing else condition for create interview block

Ehtesham599 (changes requested):
@SalShai Left a comment. Also, why are there buttons with no text and function created?

one_fm/public/js/doctype_js/job_applicant.js
SalShai: One applicant could have multiple Interviews, so thought of not including that in the condition.

SalShai:
> @SalShai Left a comment. Also, why are there buttons with no text and function created?

It's just to create a division. Cannot add a dropdown divider inside this button.

one_fm/public/js/doctype_js/job_applicant.js
```diff
@@ -739,18 +765,19 @@ var validate_min_age = function(frm) {
 };
 
 var create_interview = function(frm) {
-  frappe.route_options = {"job_applicant": frm.doc.name};
-	frappe.new_doc("Interview Result");
+	frappe.new_doc("Interview Result", {"job_applicant": frm.doc.name});
 };
-
+var check_doc_exist = function(frm, docname){
```
Ehtesham599: I think you can remove this. Everything else looks good to me.

SalShai: Done

Ehtesham599 (approved):

PR #840, opened 2022-02-15
feat:  Print Format for Contracts checkin/checkout

mymi14s:
## Feature description
A print format that adds attendance to Sales Invoice print, attendance is based on checkin/checkout time.


Ehtesham599 (approved):
LGTM. Great work!
Although, try not using joins for large datasets like attendances as it slows performance. Just use `frappe.get_list` or `frappe.get_value` for obtaining DB values. It also makes code easier to read and maintain. And since other developers use frappe database APIs instead of `frappe.sql`, it would help uniform our codebase.

PR #842, opened 2022-02-17
#181277876

SalShai:
## Feature description
- Relocate Indemnity Calculation method.
- Fix: Indemnity Allocation Calculation to exclude salary per day
- Fix: Employee Indemnity Report
- Fix: Employee Indemnity Calculation to fetch indemnity Amount instead of Basic Salary.

## Solution description
Calculate the Allocation without salary per day.

https://user-images.githubusercontent.com/29017559/154433128-1dff9466-3bbe-45c4-aef3-60c0230f25c6.mov

## Areas affected and ensured
- Employee Indemnity Calculation is based on Indemnity Amount.

  - [x] Chrome
  - [x] Safari


one_fm/one_fm/hr_utils.py
```diff
@@ -1,130 +0,0 @@
...
-def validate_leave_proof_document_requirement(doc, method):
-    '''
-        Function to validate Is Proof Document Required Flag in Leave Application
-        Triger form Validate hook of Leave Application
-    '''
-
-    if doc.leave_type and doc.status in ['Open', 'Approved']:
-        doc.is_proof_document_required = frappe.db.get_value('Leave Type', doc.leave_type, 'is_proof_document_required')
-        if doc.is_proof_document_required and not doc.proof_document:
-            frappe.throw(_("Proof Document Required for {0} Leave Type.!".format(doc.leave_type)))
```
pjamsheer: This method is not purely related to indemnity, It is related to the Leave Type and Leave Application

one_fm/hooks.py
```diff
@@ -164,7 +164,7 @@
 		"on_submit": "one_fm.utils.leave_appillication_on_submit",
 		"validate": [
 			"one_fm.utils.validate_hajj_leave",
-			"one_fm.one_fm.hr_utils.validate_leave_proof_document_requirement",
```
pjamsheer: It should be retain as it is. This is not purely related to indemnity, It is related to the Leave Type and Leave Application

one_fm/hooks.py
```diff
@@ -164,7 +164,7 @@
...
+			"one_fm.one_fm.doctype.indemnity_allocation.indemnity_allocation.daily_indemnity_allocation_builder.validate_leave_proof_document_requirement",
```
pjamsheer: Wrong path and needs to revise this change - This is not purely related to indemnity, It is related to the Leave Type and Leave Application

one_fm/one_fm/hr_utils.py
SalShai: Okay.

one_fm/hooks.py
SalShai: I'll add it back to hr_utils.py

pjamsheer (approved):

PR #864, opened 2022-02-21
chore: Check for old methods that are no longer used (especially related to Contracts)

mymi14s:
## Feature description
This PR cleaned up Contracts Doctype:

- Removed unused codes.
- Fixed the workflow which auto activates on Draft mode.
- Moved Generate Invoice and Delivery Note buttons to activate on active Contracts Only and Finance Manager Role.
- Rewrote display manager for Contracts Item based on subitem_group e.g hide site if group=service.
- Added Invoice Generator scheduler to run at 3:15am daily.
- Exported the modified Contracts workflow.


one_fm/operations/doctype/contracts/contracts.js
```diff
@@ -219,16 +219,16 @@ frappe.ui.form.on('Contracts', {
...
-		frm.fields_dict['assets'].grid.get_field('site').get_query = function() {
+		// frm.fields_dict['assets'].grid.get_field('item_code').get_query = function() {
+    //         return {
+    //             filters:{
+		// 			is_stock_item: 1,
+		// 			is_sales_item: 1,
+    //                 disabled: 0
+    //             }
+    //         }
+		// }
```
pjamsheer: Commented codes can be deleted.

mymi14s: comment removed.

pjamsheer (approved):

PR #865, opened 2022-02-22
Call gRPC external service for face recognition

Ehtesham599:
## Feature Description
Makes a call to gRPC external service instead of doing the computation locally. This eliminates the usage of `dlib` library locally to support future versions of python for the app.

## Solution description
- Implemented client stub to call external gRPC service for face recognition feature for both one_fm face_recognition page and API.
- Requires `frappe.local.conf.face_recognition_service_url` to be set in site_config. 

For reference on gRPC implementation and documentation on the request/response message, visit [this repo](https://github.com/ONE-F-M/face-recognition-service)


mymi14s (commented):
Please provide a sample use case or a method call and response sample.

Ehtesham599:
@mymi14s Check your DM. I have shared sample cases.

mymi14s (approved):
Good job

PR #868, opened 2022-02-23
chore: Extend the APIIntegration DocType to handle extra fields

mymi14s:
## Feature description
This PR adds an extra child table field in API Integration doctype to handle more API configuration key-value pair. As well as a minor fix to Pivotal Tracker that replaced the hard-coded URL with the API url.
In addition, a workspace was added for easy navigation to settings related doctypes.

image
image


Ehtesham599 (approved):
LGTM

PR #873, opened 2022-03-06
Fix: return only when shift exists

SalShai:
## Feature description
Return Shift in "get_current_shift" only if it exists/ greater than 0. This avoided fetching of an empty list.

## Solution description
- Return only when the length of the shift is greater than 0.

## Areas affected and ensured
- Fetch of the current shift of the employee.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/mobile/roster.py
+++ b/one_fm/api/mobile/roster.py
@@ -586,10 +586,6 @@ def get_current_shift(employee):
 				else:
 					if start_time <= time <= end_time:
 						return shift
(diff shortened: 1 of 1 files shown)


pjamsheer:
@SalShai I hope you checked the all the usage of this method ?!

SalShai:
> 

@pjamsheer But, it still doesn't make sense to return an empty list, right?
Though I'll look into all the usage of the function.

pjamsheer (approved):

PR #898, opened 2022-03-23
Issue Type Report

Ehtesham599:
## Feature description
As Management, I would like a report with issues list, so that we could create new issue types depending on the current problems that are arising.

## Solution description
- Created a standard script report named Issue Type.
- Added columns: Department, Issue Type, Custom Issue Type and Count.
- Default filters are user's assigned department and issue type 'Other'.

## Output screenshots
Screen Shot 2022-03-23 at 4 27 44 PM

## Areas affected and ensured
None.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/one_fm/report/issue_type/issue_type.py
+++ b/one_fm/one_fm/report/issue_type/issue_type.py
@@ -50,13 +50,3 @@ def get_data(filters=None):
 		result.append(row)
 	
 	return result
-
-@frappe.whitelist()
-def get_filtered_issue_types(doctype, txt, searchfield, start, page_len, filters):
-	res = []
-	dept = frappe.get_doc("Department", filters.get("department"))
-
-	for d in dept.issue_types:
-		res.append([d.issue_type])
-
-	return res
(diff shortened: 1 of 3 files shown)


one_fm/one_fm/report/issue_type/issue_type.py
```diff
@@ -0,0 +1,62 @@
...
+
+@frappe.whitelist()
+def get_filtered_issue_types(doctype, txt, searchfield, start, page_len, filters):
+	res = []
+	dept = frappe.get_doc("Department", filters.get("department"))
+
+	for d in dept.issue_types:
+		res.append([d.issue_type])
+
+	return res
```
pjamsheer: Instead of this you can make use of the method in `utils.py` https://github.com/ONE-F-M/One-FM/blob/2870d945a6816a9feae6d67368b341c7cf5c47dc/one_fm/utils.py#L2327 in staging branch

Ehtesham599: @pjamsheer Yeah that would be better. Thanks!

pjamsheer: Anytime

pjamsheer (approved):

PR #912, opened 2022-03-27
Hot-fix: Remove len(shift) since shift is an object/dict

SalShai:
## Feature description
Hot-fix: Remove len(shift) since shift is an object/dict

## Solution description
- Remove len(shift) since the shift is an object/dict, and not a list.
- Change time to datetime before the loop to keep it constant.

## Areas affected and ensured
- 'get_current_shift' returns shift object as usual in object form.

No.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/mobile/face_recognition.py
+++ b/one_fm/api/mobile/face_recognition.py
@@ -83,7 +83,7 @@ def verify(video, log_type, skip_attendance, latitude, longitude):
 def get_site_location(employee):
 	try:
 		shift = get_current_shift(employee)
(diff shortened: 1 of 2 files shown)


Ehtesham599:
@SalShai Looks good to me.
Since its a hotfix, change base to master-13 and request review from @taqi-onefm.

pjamsheer:
@SalShai create a branch on top of master-13 and cherry-pick the commits in this PR only and create a PR to master-13 as a hot fix.

PR #961, opened 2022-04-12
[Fix] fixing bugs reported in erp

pjamsheer:
## Feature description
- PAM Designation mandatory property fix
- Updated Designation custom fields and linking
- ERF Performance profile section pop up issue
- Job Applicant GRD section hide issue
- Remove unused codes
- Leave Application api - date formatting

  - [x] Chrome
  - [x] Safari


SalShai (approved):
LGTM

PR #977, opened 2022-04-18
#181914777: Fix Shift Assignment

SalShai:
## Feature description
Create Shift assignment for both Roster Type: Basic and Overtime.

 Since one could not have two active shift assignments, we create the one after the other. We do this by inactivating the first before creating the next.

## Solution description
Step 1: Create Shift Assignment for all the Schedule with Roster Type "Basic".
Step 2:  Fetch all the schedules with Roster type "Over-Time".
Step3: Check if the employee has a Shift Assignment for the given date, and set the status of it as "InActive".
Step4: Create Shift Assignments for the fetched schedules.

Screen Shot 2022-04-19 at 2 12 25 AM
Screen Shot 2022-04-19 at 2 12 10 AM

## Areas affected and ensured
Shift Assignment Roster Type is "Basic" is created first, and the Roster type "Over-Time" is created later.

  - [x] Chrome
  - [x] Safari

### DIFF
--- a/one_fm/api/tasks.py
+++ b/one_fm/api/tasks.py
@@ -545,21 +545,22 @@ def overtime_shift_assignment():
 	date = cstr(getdate())
 	now_time = now_datetime().strftime("%H:%M:00")
 	roster = frappe.get_all("Employee Schedule", {"date": date, "employee_availability": "Working" , "roster_type": "Over-Time"}, ["*"])
-	for schedule in roster:	
-		frappe.enqueue(process_overtime_shift,schedule=schedule, date=date, time=now_time, is_async=True, queue='long')
 
-def process_overtime_shift(schedule, date, time):
-	#Check for employee's shift assignment of the day, if he has any.
-	shift_assignment = frappe.get_doc("Shift Assignment", {"employee":schedule.employee, "start_date": date},["name","shift_type"])
-	if shift_assignment:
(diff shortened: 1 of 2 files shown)


one_fm/api/tasks.py
```diff
@@ -517,20 +517,49 @@ def end_previous_shifts():
...
+	"""
+	This method is to generate Shift Assignment for Employee Scheduling 
+	with roster type 'Over_Time'. It first looks up for Shift Assignment
+	of the employee for the day if he has any. Change the Status to "Inactive"
+	and proceeds with creating New shift Assignments with Roster Type OverTime.
+	"""
+	date = cstr(getdate())
+	now_time = now_datetime().strftime("%H:%M:00")
+	roster = frappe.get_all("Employee Schedule", {"date": date, "employee_availability": "Working" , "roster_type": "Over-Time"}, ["*"])
+	for schedule in roster:
```
Ehtesham599, line 551: Instead of iterating and then enqueuing the `process_overtime_shift` method, you can directly enqueue the method to process overtime shifts for all schedules by passing a list of schedules. This will reduce the number of background jobs created.
Modifications will have to be made to use a list instead of a schedule object in the `process_overtime_shift` method.

Ehtesham599 (changes requested):

Ehtesham599 (approved):

PR #1037, opened 2022-05-16
Feat: Enforce checkin within site geofence

Ehtesham599:
## Feature description
As a supervisor, I would like to make sure that my team checks in/out within the location geofence so I can make sure that everybody is attending at the site.

## Solution description
- Upon checkin/checkout, verify that user location is within the geofence radius set for the site.

## Areas affected and ensured
Nothing affected and ensured.

Yes. User will have to be within the geofence radius of the site while checking in/out.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
LGTM!

PR #1043, opened 2022-05-18
fix: Issue in Job Offer

pjamsheer:
## Feature description
- Issue in Job Offer - "Not allowed to change Amount after submission"

- In Job Offer we have a child table with field amount, but the amount field is setting based on the base and the salary structure selected and the field is read only too.

## Solution description
- We can set allow_on_submit true for the amount field in the child table

No

  - [x] Chrome
  - [x] Safari


SalShai (approved):
LGTM!

Ehtesham599 (approved):

PR #1074, opened 2022-05-29
Roster Projection report

Ehtesham599:
## Feature description
Generates roster projection and live projection report for sale items (service) in the contract based on employee schedules and attendances.

## Solution description
- Added a script report to view projection and live projection with chart.
- Projection is calculated based on employee schedules
- Live projection is provided until current date based on attendances.
- Report is generated for the selected contract, month and current year.

## Areas affected and ensured
Nothing affected and ensured.

No.

  - [x] Chrome
  - [x] Safari


SalShai (approved):
LGTM!

PR #1078, opened 2022-05-30
Fix: Report date value

Ehtesham599:
- Fix date value used for generating report

### DIFF
--- a/one_fm/one_fm/report/roster_projection/roster_projection.py
+++ b/one_fm/one_fm/report/roster_projection/roster_projection.py
@@ -5,13 +5,16 @@ import frappe
 from frappe import _
 import pandas as pd
 from frappe.utils import getdate, get_first_day, get_last_day
+from datetime import date
 
 def execute(filters=None):
 	if filters:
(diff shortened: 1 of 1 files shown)


one_fm/one_fm/report/roster_projection/roster_projection.py
```diff
@@ -5,13 +5,16 @@
 from frappe import _
 import pandas as pd
 from frappe.utils import getdate, get_first_day, get_last_day
+from datetime import date
 
 def execute(filters=None):
 	if filters:
-		date = str(getdate())
+		today_date = date.today()
```
mymi14s, line 12: Using system date will ignore timezone, date object with timezone can be obtained with 
from frappe.utils import nowdate
then the date object will be
getdate(nowdate())

revert if it will work for your usecase or not

Ehtesham599: The `today_date` variable is used just to get the current year for the report generation. Date inside this is ignored.

PR #1092, opened 2022-06-05
fix: Add shift Assignment permissions if missing

mymi14s:
## Feature description
On each migrate, shift assignment permission for employee is lost, this PR check if permissions exist and accordingly if required.

### DIFF
--- a/one_fm/tasks/hooks/after_migrate.py
+++ b/one_fm/tasks/hooks/after_migrate.py
@@ -1,0 +1,55 @@
+import frappe
+
+
+def execute():
+    add_shift_assignment_permissions()
+
+
+def add_shift_assignment_permissions():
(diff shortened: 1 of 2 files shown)


one_fm/tasks/hooks/after_migrate.py
```diff
@@ -0,0 +1,55 @@
...
+            }):
+            print(role)
+            perm = frappe.get_doc(shift_assignment)
+            perm.role = role
+            if role in ['HR Manager', 'HR User']:
+                perm.write = 1
+                perm.create = 1
+                perm.delete = 1
+                perm.submit = 1
+                perm.cance = 1
```
Ehtesham599, line 54: Typo here. Change to cancel.

PR #1113, opened 2022-06-12
Request Change in Site Location

SalShai:
## Feature description
- Request a shift location change by using the Shift Request form, so check-in from that particular site is allowed.
- User is allowed to have different Check In and Check Out sites.


pjamsheer:
@SalShai resolve the conflict in hooks

SalShai:
@pjamsheer , Conflict Resolved!

one_fm/public/js/doctype_js/shift_request.js
```diff
@@ -9,6 +9,7 @@ frappe.ui.form.on('Shift Request', {
 
 function set_approver(frm){
     if(frm.doc.employee){
+        console.log("Here")
```
pjamsheer, line 12: @SalShai  remove this console in next PR

pjamsheer (approved):

PR #1126, opened 2022-06-14
#182355938: Remove Authorized Signatory from Purchase Module 

SalShai:
- Remove Authorized SIgnatory Fields from all the purchase module (RFM, RFP, PO, and PR)
- Change the workflow of accepting the request.
- Remove Verification of Code during submission of the request.
- Correct few Modified Owner from doctypes.


one_fm/public/js/doctype_js/purchase_order.js
```diff
@@ -84,112 +81,6 @@ var hide_tax_fields = function(frm) {
 var hide_subscription_section = function(frm) {
 	frm.set_df_property('subscription_section', 'hidden', true);
 };
-var confirm_accept_approve_purchase_order= function(frm) {
```
pjamsheer, line 87: I hope we need OTP verification, we dont need to update the signature from employee to the PO

one_fm/public/js/doctype_js/purchase_receipt.js
```diff
@@ -8,121 +8,3 @@ this.frm.dashboard.add_transactions([
     },
 
 ]);
-
-frappe.ui.form.on('Purchase Receipt', {
-	status: function(frm){
-		confirm_accept_approve_purchase_receipt(frm);
-	}
-});
-var confirm_accept_approve_purchase_receipt = function(frm) {
```
pjamsheer, line 17: We need OTP verification, the issue is, getting the signature from employee and set to the Document

one_fm/purchase/doctype/request_for_material/request_for_material.py
```diff
@@ -58,14 +58,6 @@ def accept_approve_reject_request_for_material(self, status, reason_for_rejectio
...
 
 			if status == "Approved":
-				#fetch Signature from employee doc using user ID
-				signature = fetch_employee_signature(self.request_for_material_accepter)
-				if signature:
-					self.authority_signature = signature
-					self.save(ignore_permissions=True)
-				else:
-					frappe.throw(_("Your Signature is missing!"))
-
```
pjamsheer, line 68: This is what we need

pjamsheer:
@SalShai please check the comments

one_fm/public/js/doctype_js/purchase_receipt.js
SalShai: Okay.

one_fm/public/js/doctype_js/purchase_order.js
SalShai: @pjamsheer I believe we should remove the OTP verification because we applied the verification just so Strong authentication used for fetching the signature.

pjamsheer:
@SalShai fix the conflict

pjamsheer (approved):

PR #1142, opened 2022-06-19
fix: Shift Assignment for Active employee with filters

mymi14s:
## Feature description
Issue shift assignment for active employees only as well as apply filters for employee selection in shift assignment and employee schedule.


one_fm/one_fm/doctype/erf/erf.py
```diff
@@ -1,458 +0,0 @@
-# -*- coding: utf-8 -*-
```
SalShai: @mymi14s Was this and the above file deleted on purpose??

SalShai (approved):

PR #1179, opened 2022-06-28
HotFix: Penalty Deduction

SalShai:
- Fix: Salary Component. Change 'Basic' to 'Basic Salary'.
- Temporarily change the Calculation dates


mymi14s:
@SalShai how about making calculation date configurable in ONEFM General Setting doctype

SalShai:
> @SalShai how about making calculation date configurable in ONEFM General Setting doctype

@mymi14s yes, we could. I'm also adding the Basic Salary Component as well to be fetched from the settings.

mymi14s (approved):

PR #1252, opened 2022-07-19
Feat: Split Shift using Shift Type Doc

SalShai:
- Create Field in Shift Type to check if it's a split shift.
- Calculate the duration of the shift if it is a split shift.
- Send notification to employees to remind checkin
- send a notification to the supervisor if the employee doesn't check in.

https://user-images.githubusercontent.com/29017559/179700346-a0a29f88-6e97-4791-a3e7-639b81d91ad0.mov


one_fm/api/tasks.py
```diff
@@ -260,12 +260,8 @@ def notify_checkin_checkout_final_reminder(recipients,log_type):
 					<a class="btn btn-primary" href="/app/shift-permission/new-shift-permission-1">Submit Shift Permission</a>&nbsp;
 					""")
 	notification_category = "Attendance"
-	checkout_subject = _("Final Reminder: Please checkout in the next five minutes.")
-	checkout_message = _("""
-		<a class="btn btn-danger" href="/app/face-recognition">Check Out</a>
-		Submit a Shift Permission if you are plannig to leave early or is there any issue in checkout or forget to checkout
```
pjamsheer, line 266: @SalShai The option Submit Shift Permission needs to be there as per the requirement in checkout reminder

mymi14s (approved):

PR #1267, opened 2022-07-26
[Fix] Leave allocation builder

pjamsheer:
## Feature description
 - No leave are allocating from the scheduler since the leave allocation builder was stuck on exception

## Solution description
 - Handled the exception by try catch

reference:
Screenshot 2022-07-26 at 1 41 55 PM

## Areas affected and ensured
 - `one_fm/utils.py` L#797

No

  - [x] Chrome


SalShai (approved):
LGTM

PR #1276, opened 2022-08-01
[Fix] Ignore user permission in Accommodation Report filter

pjamsheer:
## Feature description
 - Ignore user permission in Accommodation Report filter

## Areas affected and ensured
 - `one_fm/accommodation/report/accommodation_checkin_checkout_report/accommodation_checkin_checkout_report.js`

Yes, employee filter in the accommodation checkin checkout now ignore user permission

  - [x] Chrome


SalShai (approved):
LGTM

PR #1284, opened 2022-08-03
Feat: Improve Shift Permission and Shift Request

mymi14s:
## Feature description
This PR add the following improvements:

- 1) Fix the Permissions in the Shift Permission & Shift Request Doctypes

- 2) Make Notifications for Shift Permission and Shift Request that are similar to Attendance Requests.

- 3) Shift Permission Should fetch reports to

- 4) Shift Permission & Attendance Request should have Approver Name not only ID.

# AFFECTED AREA

Workflow function previously in attendance request has been moved to one_fm.utils so as to be used globally.


one_fm/api/utils.py
```diff
@@ -11,25 +11,29 @@ def response(code, title, msg, data=None):
 
 def get_reports_to_employee_name(employee):
```
SalShai: I have a similar method in https://github.com/ONE-F-M/One-FM/blob/5d3da8da3bc07004e522e2107ccfc1cf01feeda4/one_fm/api/tasks.py#L463 I believe we should unify these methods.

mymi14s: @SalShai your method returns user_Id this one returns employee name(pk)

They are two different method

one_fm/api/utils.py
```diff
@@ -11,25 +11,29 @@ def response(code, title, msg, data=None):
...
     reports_to = frappe.db.get_value('Employee', employee, 'reports_to')
+    if reports_to:
+        return reports_to
+
     if not reports_to:
         shift = frappe.db.get_list("Shift Assignment", filters={'employee':employee},
             fields=['shift'], order_by="start_date DESC", page_length=1)
         if len(shift):
             reports_to = frappe.db.get_value("Operations Shift", shift[0].shift, 'supervisor')
+            return reports_to
```
pjamsheer: Check `if reports_to` and then return, otherwise this will return from here without checking the value is available or not.

one_fm/api/utils.py
```diff
@@ -11,25 +11,29 @@ def response(code, title, msg, data=None):
...
 
     # when no shift supervisor or reports to in employee use site and project
     if not reports_to:
         site = frappe.db.get_value('Employee', employee, 'site')
         if site:
             reports_to = frappe.db.get_value('Operations Site', site, 'account_supervisor')
+            return reports_to
```
pjamsheer: Same here too

one_fm/api/utils.py
```diff
@@ -11,25 +11,29 @@ def response(code, title, msg, data=None):
...
     if not reports_to:
         project = frappe.db.get_value('Employee', employee, 'project')
         if project:
             reports_to = frappe.db.get_value('Project', project, 'account_manager')
+            return reports_to
 
     if not reports_to:
         frappe.throw(f"Employee {employee} have no reports to.")
-
-    return reports_to
```
pjamsheer: Or we can remove the new changes of return and keep this, will give the return and the message correctly if there is no reports to

mymi14s: The function acts weird that's why there are many returns

one_fm/one_fm/custom/shift_request.json
```diff
@@ -421,9 +421,136 @@
...
+   "creation": "2018-04-13 16:32:27.974273",
+   "delete": 0,
+   "docstatus": 0,
+   "email": 1,
+   "export": 1,
+   "idx": 1,
+   "if_owner": 0,
+   "import": 0,
+   "modified": "2022-06-21 08:29:12.331484",
+   "modified_by": "s.shaikh@armor-services.com",
```
SalShai: @mymi14s Could you change these to Administrator??

SalShai (changes requested):
The 'modified_by' value needs to be changed to 'Administrator'. @mymi14s

SalShai:
@mymi14s in the conflict, kindly include both the changes.

one_fm/api/utils.py
```diff
@@ -10,6 +10,7 @@ def response(code, title, msg, data=None):
 
 
 def get_reports_to_employee_name(employee):
+    reports_to = None
```
pjamsheer: @mymi14s we can remove unused assignment

SalShai (approved):

PR #1324, opened 2022-08-21
#183040421: Payroll Report

SalShai:
Payroll Report to fetch the following Columns from all the payroll of the given month:
Employee ID, Employee Name, Project, Working Days, WP Salary, Base Salary, Civil ID, Shoon File, Bank Account												

Screen Shot 2022-08-21 at 11 54 09 AM

### DIFF
--- a/one_fm/one_fm/report/payroll_report/payroll_report.py
+++ b/one_fm/one_fm/report/payroll_report/payroll_report.py
@@ -1,0 +1,107 @@
+# Copyright (c) 2022, omar jaber and contributors
+# For license information, please see license.txt
+
+import frappe
+
+def execute(filters=None):
+	if not filters: filters = {}
+
+	if not (filters.get("month") and filters.get("year")):
+		msgprint(_("Please select month and year"), raise_exception=1)
+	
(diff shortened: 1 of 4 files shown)


one_fm/one_fm/report/payroll_report/payroll_report.py
```diff
@@ -0,0 +1,107 @@
...
+						filters.get("month"),as_dict=1)
+	
+	if not employee_details:
+		frappe.msgprint(("No Payroll Submitted this month!"), alert=True, indicator="Blue")
+	
+	for detail in employee_details:
+		row = [
+			detail.employee,
+			detail.employee_name,
+			frappe.get_value("Employee", detail.employee, ["project"]),
```
mymi14s, line 90: @SalShai say there are 1000 active employees, this will make a thousand calls, same for line 92

mymi14s (commented):
You can either join the query with `tabEmployee` or retrieve employee records, make a dict, then fetch line 90 and 92 dat from the dict. By doing the above, you'll not need to run repeated query and the report will be generated faster

PR #1326, opened 2022-08-21
chore: Exempt Attendance Request from checkin notification

mymi14s:
## Feature description
In this PR we exempt Attendance Request from checkin notification list.


SalShai (approved):

SalShai:
@mymi14s You might need to change the base.

PR #1466, opened 2022-10-31
Filter log issues via IP

NoBoneZ:
- [] Feature
- [X] Chore
- [] Bug

Only issues from the production server would show in the erpnext channel on slack

## Solution description

## Areas affected and ensured

- [X] Existing Data
- [] New Data

## Is patch required?
- [] Yes
- [] No
    ## Was the patch test?

  - [] Chrome
  - [] Safari
  - [X] Firefox

### DIFF
--- a/one_fm/tasks/erpnext/issue.py
+++ b/one_fm/tasks/erpnext/issue.py
@@ -1,4 +1,4 @@
-import socket
+import os
 
 import frappe
 from one_fm.api.slack import send_slack_message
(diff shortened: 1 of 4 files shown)


one_fm/tasks/erpnext/issue.py
```diff
@@ -1,16 +1,37 @@
...
+
+    """
+    allowed_ip = frappe.get_doc("Default API Integration").allowed_ip_address
+
+    allowed_ip_list = [i.address for i in allowed_ip]
+    
+
+    host_name = socket.gethostname()
+
+    ip = socket.gethostbyname(host_name)
```
mymi14s: @NoBoneZ this will get private ip address starting from 10.xxx.xxx.xxx not public IP

mymi14s: @NoBoneZ this will get you the actual IP address

import os
externalIP  = os.popen('curl -s ifconfig.me').readline()
print(externalIP)

one_fm/tasks/erpnext/issue.py
```diff
@@ -1,16 +1,37 @@
...
+
+
+    if ip == "34.105.145.77" or ip in allowed_ip_list:
```
mymi14s: This is a technical foul, it invalidates all algorithm above and will execute if the IP in doctype matches regardless of the host server. the number, it should be
if allowed_ip_list:

NoBoneZ:
review check

mymi14s (approved):

PR #1501, opened 2022-11-08
Allow User to attach Multiple Attachments

SalShai:
- [x] Feature
- [ ] Chore
- [ ] Bug

- This feature is to allow the user to attach multiple attachments as proof documents in the leave application.

## Solution description
- Remove the Proof Document Attach field.
- Add table consisting of attached field.
- this would allow attaching of multiple attachments.

Screen Shot 2022-11-08 at 10 11 34 AM
Screen Shot 2022-11-08 at 10 11 57 AM

## Areas affected and ensured:
- Leave application Proof Document field was removed.
- Leave application Proof Document table was added.

yes. 'Proof Document' is a table instead of a field.

- [ ] Existing Data
- [x] New Data

## Is patch required?
- [ ] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [ ] Safari
  - [ ] Firefox


pjamsheer (commented):
@SalShai Proof Documents is doctype created in onefm app, then why we have custom `one_fm/one_fm/custom/proof_documents.json`

pjamsheer (approved):

PR #1502, opened 2022-11-08
Fix: Username already exists.

SalShai:
- [ ] Feature
- [ ] Chore
- [x] Bug

- Error: Username already exists.

## Solution description
- Add Employee's full name as username.

Screen Shot 2022-11-08 at 10 47 20 AM

## Areas affected and ensured
- Username Added to user list while creating an employee.

- Username added.

- [ ] Existing Data
- [x] New Data

## Is patch required?
- [ ] Yes
- [x] No
    ## Was the patch test?

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/hiring/doctype/onboard_employee/onboard_employee.py
+++ b/one_fm/hiring/doctype/onboard_employee/onboard_employee.py
@@ -246,8 +246,11 @@ class OnboardEmployee(Document):
 			user.first_name = self.employee_name
 			user.email = self.company_email
 			user.role_profile_name = self.role_profile
-			user.save(ignore_permissions = True)
 			employee = frappe.get_doc('Employee', self.employee)
+			
+			user.username = employee.employee_name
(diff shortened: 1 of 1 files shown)


one_fm/hiring/doctype/onboard_employee/onboard_employee.py
```diff
@@ -246,8 +246,11 @@ def create_user_and_permissions(self):
 			user.first_name = self.employee_name
 			user.email = self.company_email
 			user.role_profile_name = self.role_profile
-			user.save(ignore_permissions = True)
 			employee = frappe.get_doc('Employee', self.employee)
+			
+			user.username = employee.employee_name
```
pjamsheer: @SalShai  `username` is unique, there may be chance to have same employee name for different employee.
It is better to set `employee_id` from employee as `username` in user

pjamsheer (approved):

PR #1504, opened 2022-11-08
Add Try Catch to Shift Assignment Method.

SalShai:
- [ ] Feature
- [ ] Chore
- [x] Bug

- to know if/when to shift assignment doesn’t get created

## Solution description
- Add Try Catch

## Areas affected and ensured
- Create Shift Assignment.

- No

- [x] Existing Data
- [ ] New Data

## Is patch required?
- [ ] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [ ] Safari
  - [ ] Firefox


pjamsheer (approved):

pjamsheer:
@SalShai Please fix the conflict

SalShai:
Re doing the task.

PR #1551, opened 2022-11-28
[Feat] Show stock level of item in RFM

pjamsheer:
- [x] Feature

- For the stock manager, show the warehouse wise stock level of item selected in RFM item lines

## Solution description
- Introduced a button in item lines to show stock level, the button will display only if the item code selected.

    - [x] Yes

Screenshot 2022-11-27 at 5 37 59 PM

No

- [x] Existing Data
- [x] New Data

## Is patch required?
- [x] No

  - [x] Chrome


SalShai (approved):
LGTM!

PR #1595, opened 2022-12-11
Fix: Unable to Approve Sick Leave 

SalShai:
- [ ] Feature
- [ ] Chore
- [x] Bug

- The Approver of sick leave was unable to approve the leave.
- Error: IndexError: list index out of range in update_shift_details_in_attendance
- Other Fixes in Leave Application.

## Solution description
- Check if existing and length of shift assignment.

    - [x] Yes
    - [ ] No

## Areas affected and ensured
- Leave Approver fetch
- Workflow 
- Update of attendance 

No.

- [ ] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [] No

## Is patch required?
- [ ] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [ ] Safari
  - [ ] Firefox


one_fm/api/v1/leave_application.py
```diff
@@ -329,14 +329,14 @@ def fetch_leave_approver(employee: str) -> str:
...
-    if employee_shift[0].shift:
+    if reports_to:
+        approver = frappe.get_value("Employee", reports_to, ["user_id"])
+    else if employee_shift[0].shift:
         approver, Role = get_action_user(employee,employee_shift[0].shift)
     else:
-        reports_to = frappe.get_value("Employee", employee, ["reports_to"])
-        approver = frappe.get_value("Employee", reports_to, ["user_id"])
-
+        frappe.throw(_("Please assign an approver!!"))
```
pjamsheer, line 339: If we throw from here then the approver will not be return to the api to create leave application.
If the approver returned to https://github.com/ONE-F-M/One-FM/blob/9fd9d05bc0c3698caa6dcc86e1d471ff675da276/one_fm/api/v1/leave_application.py#L265 here then only this line will execute.

SalShai: Okay.

pjamsheer:
@SalShai there is conflict too

SalShai:
Hey @pjamsheer , I'll be closing this branch and creating a new one.

PR #1633, opened 2022-12-24
feat: Email to Task Assignment

pjamsheer:
- [x] Feature

- As an Employee, I would like to send emails to ERPNext to create Tasks and assign at the same time So i can organize the coordination between external and internal parties.

## Solution description
 - Create task from email by using feature in frappe for append to document from email, then fetch the assignees from the message content. Search for tag to get the assignees and check if they exists in user then assign the task to them

    - [x] Yes

Yes, Task creation and assignment from incoming email to a specific email id

- [x] Existing Data
- [x] New Data

## Was child table created? Yes
    - Is attachment required? Yes

    - [x] No

## Is patch required?
- [x] No

  - [x] Chrome

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -368,6 +368,9 @@ doc_events = {
 	"Stock Entry": {
 		"on_submit": "one_fm.api.doc_methods.stock_entry.validate_budget"
 	},
+	"Communication": {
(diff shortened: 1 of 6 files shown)


mymi14s:
@pjamsheer this is conflict.

pjamsheer:
@mymi14s rebased

mymi14s (approved):

PR #1688, opened 2023-01-09
[CHORE]Validate job applicant doc, before creation of job offer

NoBoneZ:
- [] Feature
- [x] Chore
- [] Bug

Validation of job applicant , before the creation of job offer, to avoid errors(caused by empty mandatory fields) popping up when employee is about to created 

## Solution description

    - [] Yes
    - [x] No

## Areas affected and ensured
Job Offer

- [x] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/hiring/utils.py
+++ b/one_fm/hiring/utils.py
@@ -37,8 +37,6 @@ def get_performance_profile_guid():
 
 def validate_job_offer(doc, method):
     job_applicant = frappe.get_doc("Job Applicant", doc.job_applicant)
-    if job_applicant.status != "Accepted":
-        frappe.throw("A Job Offer cannot Be created for an applicant who is not selected")
     validate_mandatory_fields(job_applicant)
     validate_job_offer_mandatory_fields(doc)
     # Validate day off
--- a/one_fm/hiring/utils.py
+++ b/one_fm/hiring/utils.py
@@ -37,6 +37,8 @@ def get_performance_profile_guid():
 
 def validate_job_offer(doc, method):
     job_applicant = frappe.get_doc("Job Applicant", doc.job_applicant)
+    if job_applicant.status != "Accepted":
+        frappe.throw("A Job Offer cannot Be created for an applicant who is not selected")
     validate_mandatory_fields(job_applicant)
     validate_job_offer_mandatory_fields(doc)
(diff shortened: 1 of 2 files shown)


one_fm/hiring/utils.py
```diff
@@ -35,6 +36,8 @@ def get_performance_profile_guid():
         return get_url(file_path)
 
 def validate_job_offer(doc, method):
+    job_applicant = frappe.get_doc("Job Applicant", doc.job_applicant)
```
pjamsheer, line 39: We are validating the Job Applicant while we set Selected the field `one_fm_applicant_status`. So in job offer we need to validate only the value of `one_fm_applicant_status` in the Job Applicant.
1. get the `one_fm_applicant_status` from Job Applicant
2. check if the value equals to `Selected`, then proceed
3. check if the value not equals to `Selected`, then tell the user that we can create job offer for selected candidate only

NoBoneZ: @pjamsheer okay, i could add that, as further validation, but the issue was that, the absence of data in some mandatory fields of the job applicant was causing an issue, and due to the "save as draft" story we worked on for the job applicant, there is a doc.flag.ignore_validate, to make it save. Removing that(doc.flag) would solve the issue, but the "save as draft" functionality we added will not work cos of the validation

pjamsheer: @NoBoneZ So selected applicant is going to fill the applicant doc data or after fill the applicant doc data we are selecting the applicant

NoBoneZ: oh, okay okay, i will add the validation to it

mymi14s:
@pjamsheer @NoBoneZ is this good to go?

NoBoneZ:
> @pjamsheer @NoBoneZ is this good to go?

@mymi14s ,yes sir, what Mr jam spoke was already being in effect, cos selecting an applicant automatically creates a job offer, so I think it is OK sir

mymi14s (approved):

PR #2275, opened 2023-05-10
Fix: local variable 'workflow_state' referenced before assignment

SalShai:
- [ ] Feature
- [ ] Chore
- [x] Bug

- local variable 'workflow_state' referenced before the assignment
- Variable workflow_state, mandatory field needs to be defined.

## Solution description
- Define the variable.

    - [ ] Yes
    - [x] No

## Areas affected and ensured
- Email sent fine. 

- no

- [] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [] No

## Is patch required?
- [] Yes
- [] No
    ## Was the patch test?

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/processor.py
+++ b/one_fm/processor.py
@@ -11,8 +11,9 @@ def sendemail(recipients, subject, header=None, message=None,
 	sender=None, cc=None , attachments=None, delayed=False, args=None, template=None):
 	logo = "https://one-fm.com/files/ONEFM_Identity.png"
 	template = "default_email"
-	actions=pdf_link=""
+	actions=pdf_link=workflow_state=""
 	doc_link = "#"
+	mandatory_field=None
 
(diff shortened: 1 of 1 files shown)


one_fm/processor.py
```diff
@@ -11,9 +11,10 @@ def sendemail(recipients, subject, header=None, message=None,
 	sender=None, cc=None , attachments=None, delayed=False, args=None, template=None):
 	logo = "https://one-fm.com/files/ONEFM_Identity.png"
 	template = "default_email"
-	actions=pdf_link=""
+	actions=pdf_link=workflow_state=""
```
mymi14s, line 14: Please explain this line...

mymi14s (approved):

PR #2371, opened 2023-05-29
Fix: Job Applicant Notification 

SalShai:
- [ ] Feature
- [ ] Chore
- [x] Bug

- send email only when the hiring method is A la carte.
- make the message more detailed.

## Solution description
- add an if condition to check the hiring method.
- Fetch mandatory Details.
- Construct a message using mandatory details.

Screen Shot 2023-05-29 at 12 10 49 PM

## Areas affected and ensured
- Notified when A la carte.
- mandatory fields sent in email for more details.

no.

- [] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [] No

## Is patch required?
- [] Yes
- [] No
    ## Was the patch test?

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/templates/emails/default_email_with_workflow.html
+++ b/one_fm/templates/emails/default_email_with_workflow.html
@@ -44,7 +44,9 @@
                     {% endif %}
                     <br>
                     The link to the document is  <a target="_blank" class="underline" href="{{ doc_link }}"> {{reference_name}}</a> <br>
-                    Please take your action for document using the action button below.<br>
+                    {% if attachments is not none%}
+                        Please take your action for document using the action button below.<br>
+                    {% endif %}
                 </p>
                 {% for action in actions %}
                     <a href="{{ action.action_link }}" class="btn btn-primary btn-action">{{ action.action_name }}</a>
--- a/one_fm/templates/emails/default_email_with_workflow.html
+++ b/one_fm/templates/emails/default_email_with_workflow.html
@@ -51,7 +51,6 @@
                 {% for action in actions %}
                     <a href="{{ action.action_link }}" class="btn btn-primary btn-action">{{ action.action_name }}</a>
                 {% endfor %}
-                <a href="/" class="btn btn-secondary btn-action">Cancel</a>
             {% endif %}
 
             {% if content is not none %}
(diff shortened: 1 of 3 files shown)


one_fm/templates/emails/default_email_with_workflow.html
```diff
@@ -44,7 +44,9 @@ <h2 style="text-align: center;"> {{header}} </h2>
                     {% endif %}
                     <br>
                     The link to the document is  <a target="_blank" class="underline" href="{{ doc_link }}"> {{reference_name}}</a> <br>
-                    Please take your action for document using the action button below.<br>
+                    {% if attachments is not none%}
+                        Please take your action for document using the action button below.<br>
+                    {% endif %}
                 </p>
                 {% for action in actions %}
                     <a href="{{ action.action_link }}" class="btn btn-primary btn-action">{{ action.action_name }}</a>
```
pjamsheer, line 52: The Cancel button is always there can you look into it ` <a href="/" class="btn btn-secondary btn-action">Cancel</a>`

one_fm/utils.py
```diff
@@ -2593,7 +2594,6 @@ def get_mandatory_fields(doctype, doc_name):
 	if not employee_fields:
 		for link_field in meta.get_link_fields():
 			if link_field.options == 'Employee':
-				employee_fields.append(link_field.fieldname)
```
pjamsheer: Why it is removed

SalShai: Well, this was added even when no employee field exist. And it later gave me error saying the field was Nonetype.

one_fm/templates/emails/default_email_with_workflow.html
SalShai: I think it's because it exists in workflow action, no?

one_fm/utils.py
SalShai: I added it back, with some other changes.

pjamsheer: If no employee fields in mandatory fields then we fetch employee linked fields and add to the employee_fields list to set the employee data.

pjamsheer: There are cases in which the employee fields are not mandatory by doctype definition

pjamsheer (approved):

PR #2434, opened 2023-06-11
[FEATURE]Match uniform and bedding to the right ledger

NoBoneZ:
- [x] Feature
- [] Chore
- [] Bug

To match the uniform and bedding item to the correct ledgers

## Solution description
wrote validators that check the group of the items and matches them to the correct ledger, provided that there is no default ledger in the item instance

    - [x] Yes
    - [] No

## Areas affected and ensured
Stock Entry
Item

- [x] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox


Talleyrand333 (approved):

one_fm/api/doc_methods/stock_entry.py
```diff
@@ -180,4 +180,29 @@ def validate_stock_entry_items(doc, method):
...
+            account_check = {item["name"]: {"account": item["difference_account"], "group": item["subitem_group"]} for item in difference_account_field}
+            for obj in doc.items:
+                account = account_check[obj.item_code].get("account", False)
+                group = account_check[obj.item_code].get("group", False)
+                if account:
+                    obj.expense_account = account  
+                else:
+                    if group:
+                        if group == "Uniform":
+                            obj.expense_account = "Stock Issued - ONEFM"
```
pjamsheer: Hard coding is not the correct approach. The group name can be renamed, the code will not work properly.

NoBoneZ: @pjamsheer Yes, i agree, so i added a field called difference_account to the item doctype, so it fetches from that field, the only time it gets to the hard coded part was when the account was not added to the item.

PR #2639, opened 2023-08-22
feat: job offer accepted date

pjamsheer:
- [x] Feature

- Set job acceptance date in job offer

## Areas affected and ensured
- `one_fm/one_fm/custom/job_offer.json`
- `one_fm/overrides/job_offer.py`

No

- [x] Existing Data
- [x] New Data

- [x] No

## Is patch required?
- [x] No

- [x] Chrome


mymi14s (commented):
Why using one_fm_ to name, it gets more confusing.

mymi14s (approved):

PR #2653, opened 2023-08-28
[185918161]  As an HR, I want the User ID of the emp to be set as inactive when his status is changed.

Talleyrand333:
- [*] Feature
- [] Chore
- [] Bug

As an HR, I want the User ID of the emp to be set as inactive when his status is changed.

## Solution description

I Updated the employee class with the required validations

    - [] Yes
    - [*] No

## Areas affected and ensured
Employee
User

No

- [*] Existing Data
- [] New Data

## Was child table created? NO
    - [] is attachment required? NO
        did you test attachment NO
    - [] Yes
    - [*] No

## Is patch required?
- [] Yes
- [*] No
    ## Was the patch test?

  - [*] Chrome
  - [] Safari
  - [] Firefox


one_fm/overrides/employee.py
```diff
@@ -32,7 +33,22 @@ def validate(self):
                     "Employee", self.name, existing_user_id)
         employee_validate_attendance_by_timesheet(self, method=None)
         self.validate_leaves()
-        
+    
+    def update_user_doc(self):
```
NoBoneZ, line 37: @Talleyrand333 , so i was thinking, maybe you could check the previous status, cos as it looks, this db call would be made upon every single update, for example, my employee status is active, and maybe my phone number was changed or something, it makes that call to the user doctype and all, or what do you think ??

NoBoneZ (approved):

PR #2765, opened 2023-10-10
Feat: Update Holiday Attendance on update in holiday date

SalShai:
- [x] Feature
- [ ] Chore
- [ ] Bug

- When the Holiday Date is Updated, Attendance needs to be updated too.

## Solution description
- on the update of the holiday date, Find the old and new dates.
- Fix and update attendance for old date.
- and then fix and update the attendance for the new date.

    - [ ] Yes
    - [x] No

https://github.com/ONE-F-M/One-FM/assets/29017559/89ed843b-877f-4685-831e-e5573abaa4bf

## Areas affected and ensured
Holiday Attendance Marked.
Only employees which holiday attendance are considered.

no.

- [ ] Existing Data
- [x] New Data

## Was child table created?
    - [ ] is attachment required?
        did you test attachment
    - [ ] Yes
    - [x] No

## Is patch required?
- [ ] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [x] Safari
  - [x] Firefox

### DIFF
--- a/one_fm/overrides/leave_application.py
+++ b/one_fm/overrides/leave_application.py
@@ -167,7 +167,6 @@ class LeaveApplicationOverride(LeaveApplication):
             
 
     def create_or_update_attendance(self, attendance_name, date, status):
-        print(attendance_name)
         if attendance_name:
             # update existing attendance, change absent to on leave
             doc = frappe.get_doc("Attendance", attendance_name)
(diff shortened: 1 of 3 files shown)


one_fm/overrides/leave_application.py
```diff
@@ -169,7 +167,7 @@ def update_attendance(self):
             
 
     def create_or_update_attendance(self, attendance_name, date, status):
-
+        print(attendance_name)
```
pjamsheer: Remove the print

PR #2859, opened 2023-11-02
Fix: Unlink Shift assignment in Attendance Check

SalShai:
- [ ] Feature
- [x] Chore
- [ ] Bug

- issue in approving Attendance Check, since the shift is deleted.

## Solution description
Unlink Shift assignment in Attendance Check

    - [ ] Yes
    - [x] No

https://github.com/ONE-F-M/One-FM/assets/29017559/7364ca8a-2e4b-446d-a893-e9d52cda40f1

## Areas affected and ensured
Approval of Leave Application

no.

- [ ] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [ ] Yes
    - [x] No

## Is patch required?
- [ ] Yes
- [x] No
  ## Was the patch test?

- [x] Chrome
- [x] Safari
- [x] Firefox

### DIFF
--- a/one_fm/overrides/leave_application.py
+++ b/one_fm/overrides/leave_application.py
@@ -101,10 +101,9 @@ class LeaveApplicationOverride(LeaveApplication):
             frappe.log_error(frappe.get_traceback(),"Error Closing Shifts")
 
     def unlink_attendance_check(self, shift):
-        if frappe.db.exists("Attendance Check", {'shift_assignment':shift}):
-            att_check_doc = frappe.get_doc("Attendance Check", {'shift_assignment':shift})
-            att_check_doc.db_set('shift_assignment', "")
-            att_check_doc.db_set('has_shift_assignment',0)
+        attcheck_exists = frappe.db.exists("Attendance Check",  {"shift_assignment": shift})
+        if attcheck_exists:
+            frappe.db.set_value("Attendance Check", attcheck_exists, {'shift_assignment': "",'has_shift_assignment': 0})
         frappe.db.commit()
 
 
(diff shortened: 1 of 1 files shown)


one_fm/overrides/leave_application.py
```diff
@@ -90,12 +90,23 @@ def close_shifts(self):
...
+                    frappe.db.sql(query)
                 frappe.msgprint(msg = f"Shift Assignments for {self.employee_name} between {self.from_date} and {self.to_date} have been deleted",alert=1)
         except:
             frappe.log_error(frappe.get_traceback(),"Error Closing Shifts")
 
+    def unlink_attendance_check(self, shift):
+        if frappe.db.exists("Attendance Check", {'shift_assignment':shift}):
+            att_check_doc = frappe.get_doc("Attendance Check", {'shift_assignment':shift})
+            att_check_doc.db_set('shift_assignment', "")
+            att_check_doc.db_set('has_shift_assignment',0)
```
pjamsheer: This can be write like
```
a_check_exists = frappe.db.exists("A Check",  {"shift_assignment": shift})
if a_check_exists:
  frappe.db.set_value("A Check", a_check_exists, "shift_assignment", "")
  frappe.db.set_value("A Check", a_check_exists, "has_shift_assignment", 0)
```

When we do `get_doc` for just getting the values of 1 or 2 fields, a lot of unnecessary data is being fetched. This becomes worse as the number of child table in a doctype increases

`frappe.db.set_value` sets a field's value in the database, does not call the ORM triggers but updates the modified timestamp (unless specified not to).

pjamsheer: Or You can set the values like,

```
frappe.db.set_value('A Check', a_check_exists, {
    'shift_assignment': '',
    'has_shift_assignment': 0
})
```

pjamsheer (approved):

PR #3110, opened 2024-01-29
feat: employee id based on residency and employment type

pjamsheer:
- [x] Feature

- When under company residency (in Employee) is checked/unchecked, then the digit indicating residency status on the employee ID changes to 1/0
- When the Employee record is being created, then the digit indicating residency status on the employee is "S"

https://github.com/ONE-F-M/one_fm/assets/20554466/9e22f583-50ad-4aa1-a12c-b11f3672ef52

## Areas affected and ensured
- `one_fm/hiring/utils.py`
- `one_fm/overrides/employee.py`
- `one_fm/public/js/doctype_js/employee.js`

Yes, On update of residency status or employment type the system will update the employee id and notify the supervisor

- [x] Existing Data
- [x] New Data

- [x] No

## Is patch required?
- [x] No

- [x] Chrome

### DIFF
--- a/one_fm/overrides/employee.py
+++ b/one_fm/overrides/employee.py
@@ -13,6 +13,7 @@ from one_fm.hiring.utils import (
 from one_fm.processor import sendemail
 from one_fm.utils import get_domain
 from six import string_types
+from frappe import _
 
 class EmployeeOverride(EmployeeMaster):
     def validate(self):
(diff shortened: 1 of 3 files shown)


one_fm/overrides/employee.py
```diff
@@ -278,11 +299,34 @@ def notify_authorities(self):
...
-        
-        
-        
-        
-        
-    
\ No newline at end of file
+
+@frappe.whitelist()
+def get_employee_reports_to_user(employee):
```
SalShai, line 304: Hey @pjamsheer, We have a function: one_fm.api.tasks import get_action_user, which fetches the reports to an employee. Could you please check if that could be reused here?

SalShai (approved):

one_fm/overrides/employee.py
pjamsheer: In get_action_user the shift is mandatory and there are employees without shift allocation

PR #3185, opened 2024-02-18
feat: get approver of employee

pjamsheer:
- [x] Feature

Reports To Priority:
1. Set in the Employee doctype in the reports_to field  
2. The reports_to field on the Employee doctype is empty, use the supervisor field on the Operations Shift that is linked in the shift field of the Employee 
3. If no Supervisor is set for the linked Operations Shift, use the account_supervisor field on the Operations Site that is linked in the site field of the Employee  
4. If no Supervisor is set for the linked Operations Site, use the account_manager field on the Operations Site that is linked in the project field of the Employee

reports_to -> shift supervisor -> Site Supervisor(Account User) -> Site Supervisor(Account Manager)  

## Areas affected and ensured
- `one_fm/overrides/employee.py`
- `one_fm/utils.py`

- [x] Existing Data
- [x] New Data

- [x] No

## Is patch required?
- [x] No

- [x] Chrome

### DIFF
--- a/one_fm/overrides/employee.py
+++ b/one_fm/overrides/employee.py
@@ -114,7 +114,6 @@ class EmployeeOverride(EmployeeMaster):
         approver = get_approver(self.name)
         if approver:
             return frappe.db.get_value('Employee', approver, 'user_id')
-        return False
 
     def update_subcontract_onboard(self):
         subcontract_onboard = frappe.db.exists("Onboard Subcontract Employee", {"employee": self.name, "enrolled": ['!=', '1']})
--- a/one_fm/overrides/employee.py
+++ b/one_fm/overrides/employee.py
@@ -11,7 +11,7 @@ from one_fm.hiring.utils import (
     is_subcontract_employee
 )
 from one_fm.processor import sendemail
-from one_fm.utils import get_domain, get_standard_notification_template
+from one_fm.utils import get_domain, get_standard_notification_template, get_approver
 from six import string_types
 from frappe import _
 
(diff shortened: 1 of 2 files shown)


one_fm/overrides/employee.py
```diff
@@ -111,7 +111,10 @@ def inform_employee_id_update(self):
             sendemail(recipients=[reports_to], subject=subject, content=msg)
 
     def get_reports_to_user(self):
-        return get_employee_reports_to_user(self)
+        approver = get_approver(self.name)
+        if approver:
+            return frappe.db.get_value('Employee', approver, 'user_id')
+        return False
```
ks093: @pjamsheer Ideally this should return None in case no reports_to is found for an employee instead of returning False.

one_fm/utils.py
```diff
@@ -3042,31 +3042,48 @@ def has_super_user_role(user=None):
...
-    elif emp_data.reports_to:
-        return emp_data.reports_to
-    if operations_site:return operations_site
-    if operations_shift:return operations_shift
-    if not (operations_shift and operations_site and operations_shift):
-        frappe.throw("No approver found for {employee} in reports_to, site or shift".format(employee=employee))
 
+    employee_data = frappe.db.get_value('Employee', employee, ['reports_to', 'shift', 'site', 'department'], as_dict=1)
+
+    reports_to = False
```
ks093: @pjamsheer Similarly here it should be None by default instead of False.

ks093 (changes requested):

one_fm/overrides/employee.py
pjamsheer: Thank you for your suggestion buddy, I will revert it ASAP

ks093 (approved):

PR #3270, opened 2024-03-13
feat: revoked and deleted all existing oauth bearer tokens when new token is issued

ibtesam-onefm:
- [x] Feature
- [] Chore
- [] Bug

Revoke and delete all existing oauth bearer tokens for particular user when new token is issued

## Solution description
Created a doc_method and attached it with "OAuth Bearer Token" (after_insert) hook

    - [] Yes
    - [x] No

## Areas affected and ensured
OAuth Bearer Token

Yes, All the existing OAuth tokens for particular user will be deleted when new one will be issued

- [x] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/api/doc_methods/oauth_bearer_token.py
+++ b/one_fm/api/doc_methods/oauth_bearer_token.py
@@ -1,14 +1,30 @@
 import frappe
-from frappe.integrations.oauth2 import revoke_token
+import requests
+from frappe.utils import get_url
+
+
+def revoke_token(access_token):
+
+    url = f"{get_url()}/api/method/frappe.integrations.oauth2.revoke_token"
+    headers = {"Content-Type": "application/x-www-form-urlencoded"}
+    data = {"token": access_token}
+
+    try:
+        response = requests.post(url, headers=headers, data=data)
+        response.raise_for_status()
+        return True
+
+    except requests.exceptions.RequestException as e:
+        print(f"Error revoking token: {e}")
+        return False
+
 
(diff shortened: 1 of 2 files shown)


one_fm/api/doc_methods/oauth_bearer_token.py
```diff
@@ -0,0 +1,14 @@
+import frappe
+from frappe.integrations.oauth2 import revoke_token
+
+def revoke_and_delete_existing_tokens(doc, event):
+    """
+        Revokes and deletes all the existing tokens for the target user.
+    """
+    user_existing_tokens = frappe.get_all('OAuth Bearer Token', filters={'user': doc.user}, fields=['name'])
```
ks093: @ibtesam-fm get_all filters should be 
1. Client id
2. Status (get only Active tokens)
3. Additionally can even pass name filter to not include current token.

ks093 (changes requested):

one_fm/api/doc_methods/oauth_bearer_token.py
ibtesam-onefm: Sure I will do that. But we are deleting the tokens after revoking so can we neglect status filters because that seems to be useless I think?

ks093 (approved):

PR #3560, opened 2024-08-06
feat: Payroll Entry customisations

ks093:
- [x] Feature
- [] Chore
- [] Bug

1. Adding custom fields for payroll cycle start and end date in Project, Employee
2. Property setter for payroll frequency
3. Custom fields in Payroll entry for Project configuration
4. When we create a new Payroll Entry, added a project filter: All External Projects or Specific project.
5. When all external projects is selected, the system will get all the employees assigned to external projects which have the same payroll cycle as the one selected in Payroll Entry.
6. When specific project is selected, it will fetch all the employees if project's payroll cycle matches the one selected in payroll entry.

image
image

    - [x] Yes
    - [] No

- [] Existing Data
- [] New Data

  - [x] Chrome
  - [] Safari
  - [] Firefox


one_fm/fixtures/property_setter.json
```diff
@@ -1,20 +1,4 @@
 [
- {
```
NoBoneZ, line 2: You removed a lot of fields from here, was it intentional ?

ks093: Yes. Refactoring as per the new process.

one_fm/overrides/payroll_entry.py
```diff
@@ -0,0 +1,311 @@
...
+	Employee = frappe.qb.DocType("Employee")
+	Attendance = frappe.qb.DocType("Attendance")
+
+	query = (
+		frappe.qb.from_(Employee)
+		.join(SalaryStructureAssignment)
+		.on(Employee.name == SalaryStructureAssignment.employee)
+		.where(
+			(SalaryStructureAssignment.docstatus == 1)
+			& (Employee.status != "Inactive")
```
NoBoneZ: On the employee doctype, there is no inactive status

just active, courtcase, left, absconding

ks093: @AbrahamAdekunle Nice catch. Good work.

NoBoneZ (approved):

PR #3670, opened 2024-10-01
fix: Attendance Request and Shift Permission Assignment rule fixes

ks093:
- [] Feature
- [] Chore
- [x] Bug

Attendance Request - Assignment rule fixes
Shift Permission - Remove owner's assignment when moving from Draft to Pending Approver

    - [x] Yes
    - [] No

- [x] Existing Data
- [x] New Data

  - [x] Chrome
  - [] Safari
  - [] Firefox


one_fm/public/js/doctype_js/attendance_request.js
```diff
@@ -16,35 +16,9 @@ frappe.ui.form.on('Attendance Request', {
...
     }
   },
-  employee: (frm)=>{
-    // Set approver
-    frm.events.set_approver(frm);
-  },
   from_date: (frm) =>{
     validate_from_date(frm);
   },
-  set_approver: (frm) =>{
```
NoBoneZ, line 26: Please you removed the part that populates the apporove field

ks093: This is intentional. It was setting approver email in employee link field. 
Now approver details are being set in validate server side.

NoBoneZ (approved):

PR #3710, opened 2024-10-13
gender_check_for_maternity_leave

samdanikouser:
- [] Feature

        [] As OneFM
        I want to allocate maternity leave to female staff only
        so that male staff members are not accidentally allocated a maternity leave
        
        **Acceptance Criteria**
        Given that a leave allocation is being created or edited and the leave type is "Maternity Leave"
        when the user clicks save
        then the system validates that the Employee gender is female. If so the document is saved, otherwise a message is shown
        
        **Notes:**
        Maternity Leave allocation is only allowed for female workers.
        Gender is set in the Employee form

[Analyse and attach the design documentation
](https://lucid.app/lucidchart/37a0e91b-1f10-40d0-bb6c-e7126f57eb57/edit?page=dEWdTzM88_3K#)

## Solution description
Added new class LeaveAllocationOverride that overrides validation function of LeaveAllocation class from hrms application .Where this validation helps to manage the maternity leave allocation only for the female employees.

    - [] Yes

Screenshot 2024-10-13 at 3 46 54 PM

## Areas affected and ensured
leave_application.py file  - ( leave allocation feature)

 No. (Additional validation on the leave allocation)

- Yes

    - No

## Is patch required?
- No

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/overrides/leave_policy_assignment.py
+++ b/one_fm/overrides/leave_policy_assignment.py
@@ -3,6 +3,8 @@ import frappe
 from hrms.hr.doctype.leave_policy_assignment.leave_policy_assignment import (
     LeavePolicyAssignment
 )
+from hrms.hr.doctype.leave_allocation.leave_allocation import LeaveAllocation
+
 
 
 class LeavePolicyAssignmentOverride(LeavePolicyAssignment):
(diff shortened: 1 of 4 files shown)


NoBoneZ (approved):

one_fm/overrides/leave_policy_assignment.py
```diff
@@ -70,6 +72,17 @@ def create_leave_allocation(self, annual_allocation, leave_details, date_of_join
 		return allocation.name, new_leaves_allocated
 	
 
+class LeaveAllocationOverride(LeaveAllocation):
```
NoBoneZ: Kindly put this in it's own file, this file belongs to leave policy assignment

NoBoneZ (changes requested):
Kindly check the comment made

one_fm/overrides/leave_policy_assignment.py
samdanikouser: Done with the changes and updated the same as suggested.

NoBoneZ (approved):

PR #3711, opened 2024-10-14
validate maternity leave

yusuffGoodie:
- [] Feature
- [] Chore
- [] Bug

## Solution description

    - [] Yes
    - [] No

## Areas affected and ensured

- [] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [] No

## Is patch required?
- [] Yes
- [] No
    ## Was the patch test?

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -508,6 +508,7 @@ override_doctype_class = {
     "Shift Request": "one_fm.overrides.shift_request.ShiftRequestOverride",
     "Payroll Entry": "one_fm.overrides.payroll_entry.PayrollEntryOverride",
     "Salary Slip": "one_fm.overrides.salary_slip.SalarySlipOverride",
+    "Leave Allocation": "one_fm.overrides.leave_allocation.LeaveAllocationOverride",
     
     # "User": "one_fm.overrides.user.UserOverride"
 }
--- a/one_fm/overrides/leave_allocation.py
+++ b/one_fm/overrides/leave_allocation.py
@@ -1,0 +1,27 @@
+import frappe
+
+from hrms.hr.doctype.leave_allocation.leave_allocation import (
+    LeaveAllocation
+)
+from frappe import _,  bold
+
+
+class LeaveAllocationOverride(LeaveAllocation):
+	# overide default class
+
+	def validate(self):
+		super().validate()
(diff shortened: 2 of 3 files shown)


one_fm/hooks.py
```diff
@@ -508,6 +508,7 @@
     "Shift Request": "one_fm.overrides.shift_request.ShiftRequestOverride",
     "Payroll Entry": "one_fm.overrides.payroll_entry.PayrollEntryOverride",
     "Salary Slip": "one_fm.overrides.salary_slip.SalarySlipOverride",
+    "Leave Allocation": "one_fm.overrides.leave_allocation.LeaveAllocationOverride",
```
NoBoneZ, line 511: Samdani already worked on this,

one_fm/overrides/leave_allocation.py
```diff
@@ -0,0 +1,27 @@
...
+
+class LeaveAllocationOverride(LeaveAllocation):
+	# overide default class
+
+	def validate(self):
+		super().validate()
+		self.validate_maternity_leave()
+		
+	def validate_maternity_leave(self):
+		leave_type = frappe.get_doc("Leave Type", self.leave_type)
```
NoBoneZ, line 17: for future reference, the leave type on the allocation document, would be the name, so there would be no need to get the doc, then get the name, you could use it directly

one_fm/overrides/leave_allocation.py
```diff
@@ -0,0 +1,27 @@
...
+		employee = frappe.get_doc("Employee", self.employee)
```
NoBoneZ, line 18: As regards this, since you are tryng to get the gneder, you could just use the, db.get_value method, instead of taking all columns

NoBoneZ (changes requested):
KIndly check the comments, and do the needful to this PR

one_fm/overrides/leave_allocation.py
yusuffGoodie: Alright

one_fm/overrides/leave_allocation.py
yusuffGoodie: Alright

PR #3733, opened 2024-10-21
Employee status change on leave

yusuffGoodie:
- [*] Feature
- [] Chore
- [] Bug

## Solution description

    - [*] Yes
    - [] No

## Areas affected and ensured

- [*] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [*] No

## Is patch required?
- [] Yes
- [*] No
    ## Was the patch test?

  - [*] Chrome
  - [] Safari
  - [] Firefox


one_fm/utils.py
```diff
@@ -3640,4 +3640,45 @@ def send_work_anniversary_reminders():
...
+
+    # Get a list of employees with approved leave applications
+    employees_on_leave = [leave['employee'] for leave in leave_applications]
+
+    if not employees_on_leave:
+        frappe.log_error(_("No employees found with approved leave applications for today or earlier."))
+        return
+
+    # Fetch all employees with 'Active' status
+    active_employees = frappe.get_all('Employee',
```
NoBoneZ: You could filter the employees on leave in the filter, please

one_fm/utils.py
```diff
@@ -3640,4 +3640,45 @@ def send_work_anniversary_reminders():
...
+    # Fetch all employees with 'Active' status
+    active_employees = frappe.get_all('Employee', 
+        filters={
+            'status': 'Active'
+        }, 
+        fields=['name']
+    )
+
+    # Filter active employees who are in the leave applications list
+    employees_to_set_vacation = [
```
NoBoneZ, line 3671: so this would not be necessary anymore

NoBoneZ (changes requested):
@yusuffAFEX , kindly check the comments

one_fm/utils.py
yusuffGoodie: DOne

one_fm/utils.py
```diff
@@ -3640,4 +3640,38 @@ def send_work_anniversary_reminders():
...
+    current_date = today()
+
+    # Fetch approved leave applications where `from_date` is today or earlier, along with the employee's status
+    leave_applications = frappe.get_all('Leave Application', 
+        filters={
+            'status': 'Approved',
+            'from_date': ['<=', current_date],
+            'to_date': ['>', current_date]
+        },
+        fields=['employee', 'from_date', 'employee.status']  # Fetch employee status directly
```
NoBoneZ: @yusuffAFEX , you are collecting , the "from_date" field, but you are not using it, also, if that employee.status, would work, also you are not using the field

one_fm/utils.py
NoBoneZ: also, you still need to filter out inactive employees, 

so you need to get a list of name(id) of employee on leave, then also those that are active,

yusuffGoodie: I don't understand this. Why do I need filter out inactive

NoBoneZ: @yusuffAFEX , so it is possible to have  employees with a different status from active, and the aim of this method is to set those with active status, to vacation, and let's assume, an employee is already set to vacation,  there would be no need to  set it again

one_fm/utils.py
yusuffGoodie: Okay

one_fm/utils.py
yusuffGoodie: Yes, that's exactly what the list comprehension is doing. It is filter those who that have active status to set them to vacation.

NoBoneZ: so, here, it is leave status that is being checked, not the employee status
image

yusuffGoodie: It is employee status. I tested it

NoBoneZ: Oh, great great, thank you very much

NoBoneZ (approved):

PR #3774, opened 2024-11-06
I want to set Employee status as "Active" if Leave Application is cancelled after the leave start date

samdanikouser:
- [] Feature

https://github.com/user-attachments/assets/5ed1bd75-e7d0-4b65-9a22-b74a9c537bb2

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/public/js/doctype_js/leave_application.js
+++ b/one_fm/public/js/doctype_js/leave_application.js
@@ -32,6 +32,10 @@ frappe.ui.form.on("Leave Application", {
                             } else {
                                 frappe.dom.freeze();
                                 frm.set_value('custom_reason_for_cancel', values.reason);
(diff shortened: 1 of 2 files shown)


one_fm/public/js/doctype_js/leave_application.js
```diff
@@ -32,6 +32,10 @@ frappe.ui.form.on("Leave Application", {
                             } else {
                                 frappe.dom.freeze();
                                 frm.set_value('custom_reason_for_cancel', values.reason);
+                                const today = new Intl.DateTimeFormat('en-CA').format(new Date());
```
NoBoneZ: please for today, you can use this
frappe.datetime.now_date()

NoBoneZ (approved):

PR #3799, opened 2024-11-18
Feat : As OneFM, I want the Leave Application new date proposal  notification sent to the Employee to follow Pathfinder format

samdanikouser:
- [] Feature

https://github.com/user-attachments/assets/1e9c556f-7b8d-4017-aab2-85408f10bd2e


one_fm/overrides/leave_application.py
```diff
@@ -644,29 +644,33 @@ def reassign_to_applicant(employee: str, leave_name: str):
 
 @frappe.whitelist()
 def send_proposed_date_email(doc_name):
-    doc = frappe.get_doc("Leave Application", doc_name)
     frappe.db.set_value("Leave Application",doc_name,'workflow_state',"New Dates Proposed")
```
NoBoneZ, line 637: If you are going to do this, you can also use the apply workflow method, just in case there are assignment rules or stuffs like that

one_fm/overrides/leave_application.py
```diff
@@ -644,29 +644,33 @@ def reassign_to_applicant(employee: str, leave_name: str):
...
     args = frappe._dict({
-                    "employee_name_eng" : doc.employee_name,
-                    "employee_name_arabic" : employee_info[0].get("employee_name_arabic"),     
-                    "employee_id" : doc.employee,
-                    "leave_type_eng" : doc.leave_type,
-                    "start_data" : doc.from_date,
-                    "end_date" : doc.to_date,
-                    "total_days" : doc.total_leave_days,
-                    "date_of_application" : doc.posting_date,
+                    "employee_name_in_arabic": employee[0].employee_name_in_arabic,
```
NoBoneZ, line 641: please kindly check this, it should return as a dict, not an array of dicts, please check this again

one_fm/overrides/leave_application.py
samdanikouser: yes right now , the proposed dates are implemented independently because of this its implemented in this way . to make the proposed dates in workflow its another chore that i am currently working on , i shall remove this update once on doing that .

one_fm/overrides/leave_application.py
samdanikouser: Hi @Abraham Adekunle , the received data is in tuple DT ({'employee_name_in_arabic': 'مشعل حمدان'},) , so to access this i am using index values 
data = ({'employee_name_in_arabic': 'مشعل حمدان'},) 
employee_dict = data[0]

in this way 
we can also access it by the key 
    employee_name = employee_dict['employee_name_in_arabic']
after checking if 
if 'employee_name_in_arabic' in employee_dict:
employee_name = employee_dict['employee_name_in_arabic']
 
 which is the suggested way ?

NoBoneZ (approved):

PR #3817, opened 2024-11-24
BUG: ISSUE WITH DAILY ROSTER PROJECTION VIEW

NoBoneZ:
- [] Feature
- [] Chore
- [X] Bug

There was an undefined method causing the algorithim to break

## Solution description
The method was removed from frappe, i checked for its replacement, but could not find any, i added it to one_fm utils

    - [] Yes
    - x[] No

## Areas affected and ensured
Daily Roster Projection VIew

- [x] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/tasks/one_fm/daily.py
+++ b/one_fm/tasks/one_fm/daily.py
@@ -5,6 +5,7 @@ from frappe.utils import getdate, get_last_day, add_days, today
 from frappe.utils.data import get_url_to_form
 from frappe.desk.query_report import *
 from one_fm.processor import sendemail
+from one_fm.utils import background_enqueue_run
 
 
 def generate_contracts_invoice():
--- a/one_fm/utils.py
+++ b/one_fm/utils.py
@@ -21,6 +21,7 @@ from frappe.utils.user import get_users_with_role
 from frappe.permissions import has_permission
 from frappe.model.workflow import apply_workflow
 from frappe.desk.form import assign_to
+from frappe.desk.query_report import get_report_doc
 from frappe.model.naming import make_autoname
 from deep_translator import GoogleTranslator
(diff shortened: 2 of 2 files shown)


one_fm/tasks/one_fm/daily.py
```diff
@@ -5,6 +5,7 @@
 from frappe.utils.data import get_url_to_form
 from frappe.desk.query_report import *
 from one_fm.processor import sendemail
+from one_fm.utils import background_enqueue_run
```
pjamsheer, line 8: Why this import here

one_fm/utils.py
```diff
@@ -3644,4 +3645,28 @@ def send_work_anniversary_reminders():
                     others = [d for d in anniversary_persons if d != person]
                     reminder_text = get_work_anniversary_reminder_text(others)
                     send_work_anniversary_reminder(person_email, reminder_text, others, message, sender)
+
+
+def background_enqueue_run(report_name, filters=None, user=None):
```
pjamsheer, line 3650: Where this method is using

one_fm/tasks/one_fm/daily.py
NoBoneZ: The method was a frappe method, but it was removed, and there was no replacement
@pjamsheer

one_fm/utils.py
NoBoneZ: This method is the one that was removed by frappe

pjamsheer (approved):

PR #3865, opened 2024-12-15
Chore : Changing the on_cancel hook to the class function as both the hook and the class function is defined in present

samdanikouser:
- [] Chore

## Clearly and concisely describe the chore .
Chore : Changing the on_cancel hook to the class function as both the hook and the class function is defined in present

Changing the on_cancel hook to the class function as both the hook and the class function is defined in present

## Solution description
leave_application.py file changes from hook.py file

    - [] No

https://github.com/user-attachments/assets/bf76defe-ac27-4613-8d3f-294482438ce3

## Areas affected and ensured
leave_application.py

No. 

- [] Existing Data

## Was child table created?
    - No
    
    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/overrides/leave_application.py
+++ b/one_fm/overrides/leave_application.py
@@ -431,7 +431,8 @@ class LeaveApplicationOverride(LeaveApplication):
             reassign_responsiobility.reassign()
         self.create_leave_ledger_entry(submit=False)
         # notify leave applier about cancellation
-        leave_application_on_cancel()
+        method = "on_cancel"
+        leave_application_on_cancel(self,method)
         self.cancel_attendance()
         self.validate_cancel()
         send_leave_cancellation_email_to_leave_approver(self)
--- a/one_fm/overrides/leave_application.py
+++ b/one_fm/overrides/leave_application.py
@@ -431,8 +431,7 @@ class LeaveApplicationOverride(LeaveApplication):
             reassign_responsiobility.reassign()
         self.create_leave_ledger_entry(submit=False)
         # notify leave applier about cancellation
-        method = "on_cancel"
-        leave_application_on_cancel(self,method)
+        leave_application_on_cancel(self,"on_cancel")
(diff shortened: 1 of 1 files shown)


one_fm/overrides/leave_application.py
```diff
@@ -431,7 +431,8 @@ def on_cancel(self):
             reassign_responsiobility.reassign()
         self.create_leave_ledger_entry(submit=False)
         # notify leave applier about cancellation
-        leave_application_on_cancel()
+        method = "on_cancel"
```
NoBoneZ: why do you need to declare this particular variable here

samdanikouser: The leave_application_on_cancel() function was previously invoked via a hook, where we performed the Hajj check for the leave application. This check happens in two cases:

When cancelling the leave.
When submitting the leave.
The hajj_check() function takes two parameters: doc (the leave application document) and method (the action being performed, such as "submit" or "cancel"). To differentiate between the cancellation and submission scenarios, we explicitly pass the method (cancel or submit) to the hajj_check() function.

Here’s how we ensure that the Hajj check is executed correctly for both actions:

When cancelling a leave, we set the method parameter to "on_cancel".
When submitting a leave, we set the method parameter to "on_submit".
This way, the hajj_check() function knows which action is being performed, allowing it to run the appropriate logic for each case.

NoBoneZ: @samdanikouser , if that be the case, for this, eave_application_on_cancel(self,method)


you can just do this, eave_application_on_cancel(self,"on cancel"), instead of declaring another variable, or is it being used in multiple places

samdanikouser: Notes. Changes done !

NoBoneZ (approved):

PR #3960, opened 2025-01-23
FEAT : As an ERP user,   I want the ToDo references(Links) in the Google Task when I create a Todo in the ERP,   so that I can action the ToDo from the Google Task

samdanikouser:
- [] Feature

https://www.pivotaltracker.com/story/show/188807599

(https://www.pivotaltracker.com/story/show/188807599)

## Solution description
Adding the reference details in the notes of google task
The reference links are added as url in the notes as the google task right now dont support any html features.When we create a task with the Gmail email link, Google Tasks will display it as a clickable link in the notes field. However, the "button next to the date" behavior is reserved for tasks created directly from Gmail, as it's part of the native integration between Gmail and Google Tasks

    - [] No

## Output screenshots 
Screenshot 2025-01-23 at 1 06 01 PM

## Areas affected and ensured
Todo.py file

 No.

- [] Existing Data

## Was child table created?
    -No

    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox


one_fm/overrides/todo.py
```diff
@@ -110,19 +110,49 @@ def create_google_task_on_todo_creation_in_erp(doc, method):
...
+            table_content = "\n".join(
+                f"{' : '.join(td.get_text(strip=True) for td in row.find_all('td'))}" for row in rows
+            )
+            text_output += "\n\nDetails:\n" + table_content +"\n"
+
+        return text_output
     except Exception as e:
-        frappe.throw(_("Error while checking tasks: {e}"))
-    return False
+        print(f"Error converting HTML to plain text: {e}")
```
ks093: @samdanikouser Add frappe.throw or frappe.log_error instead of printing the error.

ks093 (approved):

PR #3999, opened 2025-02-09
FEAT : As an Employee, I want to set my reliever in the leave application from the mobile app, So that the reliever gets assigned all my tasks and todos during my leave period.

samdanikouser:
- [] Feature

https://www.pivotaltracker.com/story/show/188861486

    - [] No

Screenshot 2025-02-09 at 4 29 52 PM

Screenshot 2025-02-09 at 4 29 59 PM

## Areas affected and ensured
None

No. 

- [] Existing Data

## Was child table created?
    - No
    
    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/api/v1/leave_application.py
+++ b/one_fm/api/v1/leave_application.py
@@ -207,10 +207,18 @@ def get_leave_types(employee_id: str = None) -> dict:
         return response("Internal Server Error", 500, None, error)
 
 
+@frappe.whitelist()
+def get_employees_list():
+    employees = frappe.get_all("Employee",fields=["employee","employee_name","employee_id"])
+    employee_list = {}
+    for employee in employees:
+        employee_list[employee.employee] = f'{employee.employee_name}-[{employee.employee} ]-[{employee.employee_id}]'
+    return response("Success", 200, employee_list)
+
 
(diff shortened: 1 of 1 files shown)


one_fm/api/v1/leave_application.py
```diff
@@ -207,10 +207,18 @@ def get_leave_types(employee_id: str = None) -> dict:
         return response("Internal Server Error", 500, None, error)
 
 
+@frappe.whitelist()
+def get_employees_list():
```
pjamsheer, line 211: What is the use of this method

ibtesam-onefm: This is being used to fetch employees to show as reliver field options in ionic app leave application form. I think we should add a filter (is_reliver -> true) in it.

samdanikouser: Please correct me ,according to my understanding from Ebenezer ,The reliever field in the employee doctype is related to roster users ,That is not for the leave application .  
i missed to add the filter of active status . i shall add and push the changes .

ibtesam-onefm (approved):

PR #4083, opened 2025-03-16
FEAT : [Susan] As a Interviewer  I want to keep the From Time and To Time as Non Mandatory for A La Carte Recruitment.  so that I can move on to next step without filing it

samdanikouser:
- [] Feature

https://www.pivotaltracker.com/story/show/188905019

## Solution description
Interview.JSON , interview.py file changes

    - [] No

https://github.com/user-attachments/assets/5e2fca0e-bbf5-415f-8ff5-ba7c5e738b56

## Areas affected and ensured
Interview list file 

No.

- [] Existing Data

## Was child table created?
    - No
    
    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox


one_fm/overrides/interview.py
```diff
@@ -49,6 +49,10 @@ def update_interview_rounds_in_job_applicant(doc, method):
 
 
 class InterviewOverride(Interview):
+    def validate(self):
```
NoBoneZ: Oh this would not work, they do not want it to mandatory, this would always set the time to the assigned time here


in as much as they do not want it ot be mandatory, they also do not want a prefilled value, and they might want to assign a time to it later on

@samdanikouser

samdanikouser: @NoBoneZ  Since the From Time and To Time fields are mandatory in the HR DocType (which is not a custom DocType in one_fm), we cannot set them to null from the custom form. To handle this, I have added an after_insert hook to satisfy the requirement of setting the From Time and To Time fields to null when the Hiring Method is "A la carte Recruitment".

NoBoneZ: @samdanikouser , yes i understand, but the reason why that can not work is that it would always be none, eben if they want to set it, it would be none, and it can be made non mandatory, if you use property setter, and you export the fixtures

one_fm/hooks.py
```diff
@@ -376,6 +376,7 @@
 	},
 	"Interview": {
 		"validate": "one_fm.overrides.interview.update_interview_rounds_in_job_applicant",
+        "after_insert": "one_fm.overrides.interview.update_from_to_date_null",
```
NoBoneZ, line 379: This can not work because, it would always default the value to None, so the user wants to have the flexibility of not needing to set a value, and it would be empty, if we do it from the backend, if the iuser wants to set a date, it would be set as None

@samdanikouser

one_fm/overrides/interview.py
samdanikouser: @NoBoneZ  , Thankyou for the suggestions i have made the appropriate changes and updated the PR

one_fm/hooks.py
samdanikouser: The after_insert hook will be updating the values only at the time of insertion , therefore user will not face issue with updating the values in future if required .

NoBoneZ (approved):

PR #4151, opened 2025-04-03
hot-fix: added batching for post editing in roster

ibtesam-onefm:
- [] Feature
- [] Chore
- [x] Bug

Requests times out while scheduling posts in Roster

## Solution description
- Added batching for all types of roster edits (Plan Post, Cancel Post, Suspend Post, Post Off)

    - [] Yes
    - [x] No

## Areas affected and ensured
Roster (Post View)

No

- [x] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/one_fm/page/roster/roster.py
+++ b/one_fm/one_fm/page/roster/roster.py
@@ -824,45 +824,51 @@ def edit_post(posts, values):
         frappe.throw(_("Insufficient permissions to Edit Post."))
 
     args = frappe._dict(json.loads(values))
+    posts = json.loads(posts)
 
-    if args.post_status == "Plan Post":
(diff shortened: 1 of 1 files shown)


ks093:
@ibtesam-onefm @pjamsheer While this should work fine, but the end user will not be able to know if the posts have been edited or not since we are doing this in background. 
We probably need to do something similar to what we have done in dayoff function in the same file for giving instant response.

pjamsheer (approved):

PR #4167, opened 2025-04-09
FEAT : [Abdullah] As OneFM  I want to auto-set the Completed By and Completed on fields on the task doctype when the status changes to "Pending Review" so that a user does not have to set it every time

samdanikouser:
- [] Feature

https://www.pivotaltracker.com/story/show/189044952

    - [] No

https://github.com/user-attachments/assets/f9b3851c-d3e8-4925-901f-86366aa1df62

## Areas affected and ensured
Task doctype

No.

- [] Existing Data

## Was child table created?
    - None
    
    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox


one_fm/overrides/task.py
```diff
@@ -22,6 +23,17 @@ def validate_task(doc, method):
 
 def after_task_insert(doc, method):
     sync_assign_to_field(doc)
+    check_completed_by_and_completed_on(doc,method)
+
+
+def check_completed_by_and_completed_on(doc, method):
+    if doc.status == "Pending Review" or doc.status=="Completed":
+        if not doc.completed_on or doc.completed_on != frappe.utils.nowdate():
```
ibtesam-onefm, line 30: This will update the completed_on field when status is changed to "Completed". According to story we just want to capture the date when status is updated to "Pending Review"

one_fm/public/js/doctype_js/task.js
```diff
@@ -9,6 +9,20 @@ const USER_PERMS = {
 }
 
 frappe.ui.form.on("Task", {
+    status(frm) {
```
ibtesam-onefm: I would suggest to keep things conventional. Use like following

status: function (frm) {

ibtesam-onefm (approved):

PR #4181, opened 2025-04-13
feat: add restriction to google task updation if linked to doc in ERP

ibtesam-onefm:
- [x] Feature
- [] Chore
- [] Bug

As OneFM
I want to restrict users from closing ERPNext todos associated with other doctype records
so that the required action is always captured on the respective doctype records

## Solution description
Re-opened task with additional restriction details (If task linked to doc in ERP is updated from Google Tasks)

    - [] Yes
    - [x] No

## Areas affected and ensured
ToDo

No

- [x] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/overrides/todo.py
+++ b/one_fm/overrides/todo.py
@@ -90,6 +90,11 @@ def create_google_task_on_todo_creation(doc, method):
 
 def create_google_task_on_todo_creation_in_erp(doc, method):
     employee_email = doc.allocated_to
+    
+    # If task is already created then avoiding recreating it
+    if doc.custom_google_task_id:
+        return
+    
     if not employee_email:
         frappe.throw(_("No assigned user found for this ToDo"))
     service = get_google_task_service(employee_email)
@@ -131,6 +136,8 @@ def create_description_for_google_todo(doc):
         todo_doc_type = doc.reference_type if doc.reference_type else ""
         todo_reference_link = get_url_to_form(todo_doc_type, doc.reference_name) if doc.reference_name else ""
         task_notes +=f"""
+        Hey you cant update this task on Google Task, Close the task in ERPNext
+        
 		ToDo Reference: {todo_reference}
 		Reference DocType: {todo_doc_type}
 		Reference Name: {todo_reference_link}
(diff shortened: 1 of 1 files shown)


one_fm/overrides/todo.py
```diff
@@ -269,24 +276,36 @@ def sync_google_tasks_for_users(user_emails=[]):
...
+                todo = frappe.get_doc('ToDo', todos[0]['name'])
                 todo.db_set("description", task_description)
-                todo.db_set("status", mapped_status)
                 todo.db_set("custom_google_task_title", task_title)
                 todo.db_set("date", due_date)
                 todo.db_set("allocated_to", allocated_to)
-                todo.db_set("custom_source", custom_source)
+
+                # If status doesnot match
+                if mapped_status != todo.status:
```
samdanikouser, line 297: @ibtesam-onefm  Why are we checking the status twice? We're already checking if todo.reference_type and mapped_status == 'Closed' in the next line, which is when we need to reopen the task that is closed.

one_fm/overrides/todo.py
```diff
@@ -90,6 +90,11 @@ def create_google_task_on_todo_creation(doc, method):
 
 def create_google_task_on_todo_creation_in_erp(doc, method):
     employee_email = doc.allocated_to
+    
+    # If task is already created then avoiding recreating it
+    if doc.custom_google_task_id:
```
samdanikouser, line 95: @ibtesam-onefm What will happen if the task is deleted from the google task?

ibtesam-onefm: Deleting a task on Google Task will not remove the id from ToDo so this would prevent redundancy of task creation process

one_fm/overrides/todo.py
ibtesam-onefm: This condition is added for all the tasks that have conflicting status. In the next check we are adding the required restriction. This would be beneficial when we will require to add some more status based restrictions

pjamsheer (approved):

PR #4217, opened 2025-04-22
Feat : As OneFM I want to auto-fetch All non-external project users in to the MOM General Attendance table so that their Attendance can be marked

samdanikouser:
- [] Feature

https://www.pivotaltracker.com/story/show/189075666

    - [] No

https://github.com/user-attachments/assets/26b73a73-9db2-429a-a338-7c5236af1001

## Areas affected and ensured
MOM Creation

No

- [] Existing Data

## Was child table created?
    - None
    
    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox


pjamsheer:
@samdanikouser change the base to staging

Talleyrand333 (approved):

PR #4240, opened 2025-04-28
FEAT : As OneFM I want the Attendance Request with Workflow State Pending Approval to be approved automatically right before the Attendance Marking so that Employees with Valid Attendance Requests are not marked Absent

samdanikouser:
- [] Feature

https://one-fm.atlassian.net/browse/ON-5

    - [] No

## Areas affected and ensured
None

No

- [] Existing Data

## Was child table created?
    -None

    - [] No

## Is patch required?
- [] No

  - [] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -705,7 +705,6 @@ scheduler_events = {
 			'one_fm.api.tasks.generate_payroll'
 		],
 		"05 23 1-31 * *": [ #“At 23:05 on every day-of-month from 1 through 31.”
-			# 'one_fm.overrides.attendance_request.mark_future_attendance_request',
 			'one_fm.tasks.one_fm.daily.roster_projection_view_task',
 		],
 		"15 0 * * *": [ # create shift assignment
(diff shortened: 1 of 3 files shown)


one_fm/hooks.py
```diff
@@ -705,7 +705,7 @@
 			'one_fm.api.tasks.generate_payroll'
 		],
 		"05 23 1-31 * *": [ #“At 23:05 on every day-of-month from 1 through 31.”
-			'one_fm.overrides.attendance_request.mark_future_attendance_request',
+			# 'one_fm.overrides.attendance_request.mark_future_attendance_request',
```
pjamsheer: Why this is commented

samdanikouser: This is the cron that runs at 11:05 for marking , just commented it for future reference .

samdanikouser: @pjamsheer Removed the commented cron job.

pjamsheer (approved):

PR #4320, opened 2025-05-12
fix: (ON-41) Days off notification issue

yusuffGoodie:
- [] Feature
- [] Chore
- [x] Bug

https://one-fm.com/app/hd-ticket/838

## Solution description

    - [x] Yes
    - [] No

Currently, the query is grouped by shift. So, if an employee is rostered for day-off OT on two different shift, it returns something like [{ 'ot_days': 1 }, { 'ot_days': 1 }]. However, only the first item is used, which causes the supervisor to keep receiving day-off notifications even after assigning day-off OT to the employee.

I removed the GROUP BY clause so that the query returns a combined result like [{ 'ot_days': 2 }].

## Areas affected and ensured

- [] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/operations/doctype/roster_day_off_checker/roster_day_off_checker.py
+++ b/one_fm/operations/doctype/roster_day_off_checker/roster_day_off_checker.py
@@ -114,7 +114,8 @@ def validate_offs(employee):
 		# Calculate no of off days
 		od = frappe.db.sql(frappe.qb.from_(EmployeeSchedule)        
 			.select(Count("name").as_("off_days"))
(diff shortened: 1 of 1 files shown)


ks093 (dismissed):
@yusuffGoodie Please add more info in PR. 
1. What was cause of the issue?
2. What has been done to fix it?

ks093 (approved):

PR #4405, opened 2025-06-01
fix: Employee Checkin Notification Issue

yusuffGoodie:
- [] Feature
- [] Chore
- [x] Bug

https://one-fm.com/app/hd-ticket/877

## Solution description

For the first case, I discovered that the Employee Check-in notification was using the "Supervisor Reminder Shift Start" time instead of the "Late Entry Grace Period." This caused the notification to be sent after the "Supervisor Reminder Shift Start" time, even though the employee checked in between the "Late Entry Grace Period" and the "Supervisor Reminder Shift Start" time. As a result, the supervisor did not receive the notification. I have made the necessary changes to use the "Late Entry Grace Period" instead.

For the second case, I discovered that the shift permission filter was incorrect. It was excluding all employees with shift permission, without excluding only those with rejected shift permissions. I have corrected the filter accordingly.

    - [x] Yes
    - [] No

## Areas affected and ensured

- [] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/api/tasks.py
+++ b/one_fm/api/tasks.py
@@ -1773,7 +1773,7 @@ def fetch_employees_not_in_checkin():
 		shift_permissions = [i.employee for i in frappe.db.sql(f"""
 			SELECT employee FROM `tabShift Permission`
 			WHERE date='{cur_date}' AND employee IN {shift_assignments_employees_tuple}
-			AND log_type='{log_type}'
+			AND log_type='{log_type}' AND workflow_state != 'Rejected'
 			GROUP BY employee
 		""", as_dict=1)]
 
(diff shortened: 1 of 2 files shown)


one_fm/api/tasks.py
```diff
@@ -1785,6 +1785,7 @@ def fetch_employees_not_in_checkin():
 			WHERE  docstatus=1
 			AND '{cur_date}' BETWEEN from_date AND to_date
 			AND employee IN {shift_assignments_employees_tuple}
+			AND workflow_state NOT IN ('Rejected', 'Canceled')
```
Talleyrand333, line 1788: @yusuffGoodie no workflow state in attendance request called "Canceled"

yusuffGoodie: I saw it for attendance request.

image

Talleyrand333 (approved):

PR #4460, opened 2025-06-17
feat: batch wise post schedule checker

pjamsheer:
- [x] Feature

- Batching the Post Schedule Checker creation

## Solution description
- Enqueued the method for batches of 10 contracts each

## Areas affected and ensured
- `one_fm/hooks.py`
- `one_fm/operations/doctype/post_scheduler_checker/post_scheduler_checker.py`

No

- [x] Existing Data
- [x] New Data

- [x] No

## Is patch required?
- [x] No

- [x] Chrome

### DIFF
--- a/one_fm/operations/doctype/post_scheduler_checker/post_scheduler_checker.py
+++ b/one_fm/operations/doctype/post_scheduler_checker/post_scheduler_checker.py
@@ -7,6 +7,7 @@ from frappe.utils import getdate, get_last_day, get_first_day, date_diff
 from frappe.model.document import Document
 from one_fm.utils import get_week_start_end
 from one_fm.operations.doctype.operations_shift.operations_shift import get_supervisor_operations_shifts, get_shift_supervisor
+import math
 
 class PostSchedulerChecker(Document):
 	def autoname(self):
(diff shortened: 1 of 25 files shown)


one_fm/operations/doctype/post_scheduler_checker/post_scheduler_checker.py
```diff
@@ -172,4 +173,40 @@ def get_post_schedules(project, post, first_day, last_day):
...
+				`tabContracts` c JOIN `tabProject` p ON p.name = c.project
+			WHERE
+				c.workflow_state = 'Active' AND p.is_active = 'Yes'
+	""")[0][0]
+
+	if count:
+		page = 1
+		page_size = 10
+		iterations = math.ceil (count / page_size)
+		for page in range(page, iterations):
```
ibtesam-onefm: Suggestion: I think its better not to reuse the same variable (as declared before loop). Instead of "page" it can be "current_page" or something like this just to improve readability

ibtesam-onefm: Please don't forget to test your for loop. I think it should be **range(page, iterations + 1)** because range doesn't include the ending number

PR #4528, opened 2025-07-08
Allow post overfill in Roster

ks093:
- [x] Feature
- [] Chore
- [] Bug

Before: Overfilling a post was not allowed in Roster. It wasn't working properly as intended.
Now: Roster will schedule staff even if post is being overfilled. A warning message will be shown with the Role, Shift and dates on which overfill has happened.  

image

- [x] Existing Data
- [x] New Data

## Is patch required?
- [] Yes
- [x] No

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/one_fm/page/roster/roster.py
+++ b/one_fm/one_fm/page/roster/roster.py
@@ -574,15 +574,7 @@ def extreme_schedule(employees, shift, operations_role, otRoster, start_date, en
 			`operations_role`, `post_abbrv`, `roster_type`, `day_off_ot`, `start_datetime`, `end_datetime`, `owner`, `modified_by`, `creation`, `modified`)
 			VALUES
 		"""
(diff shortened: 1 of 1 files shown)


one_fm/one_fm/page/roster/roster.py
```diff
@@ -742,7 +740,8 @@ def validate_overfilled_post(date_list_pd, operations_shift_name):
 		schedule_dict[each_item.get("date").strftime("%Y-%m-%d")] = each_item.schedule_count
 
 
-	return{"schedule_dict":schedule_dict,"post_number":post_number}
+	print("schedule_dict", schedule_dict, "post_number", post_number)
```
NoBoneZ: Please, remove print statement
@ks093

NoBoneZ (approved):

PR #4680, opened 2025-08-17
As OneFM I want to notify the Reports To/Manager if their subordinates fail to submit their Employee Daily Action for the previous day so that they can follow up with them

NoBoneZ:
- [x] Feature
- [] Chore
- [] Bug

To send notifications to the reports to when their reportee fails to submit the employee daily action

## Solution description
Added a class that handles the logic, for checking the neccessary conditions

    - [] Yes
    - [] No

## Areas affected and ensured
EMployee Daily Action

- [x] Existing Data
- [x] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [x] Yes
- [] No
    ## Was the patch test?
    
image

  - [x] Chrome
  - [] Safari
  - [] Firefox

### DIFF
--- a/one_fm/hooks.py
+++ b/one_fm/hooks.py
@@ -634,9 +634,6 @@ scheduler_events = {
 		"45 10 * * *": [
 			'one_fm.utils.send_travel_agent_email'
 		],
-        "00 4 * * *": [ 
-			"one_fm.one_fm.doctype.employee_daily_action.employee_daily_action.run_employee_daily_action_check_notifications"
-		],
 		"10 4 * * *": [ #“At 04:10.”
 			'one_fm.utils.check_grp_operator_submission_four',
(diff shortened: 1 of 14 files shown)


one_fm/hooks.py
```diff
@@ -634,6 +634,9 @@
 		"45 10 * * *": [
 			'one_fm.utils.send_travel_agent_email'
 		],
+        "00 4 * * *": [ 
+			"one_fm.one_fm.doctype.employee_daily_action.employee_daily_action.run_employee_daily_action_check_notifications"
+		],
```
pjamsheer: Need to move this into Process Task

NoBoneZ: @pjamsheer , Done

Talleyrand333 (approved):

PR #5714, opened 2026-03-09
[FEAT][Workflow for onefm developer v15]

samdanikouser:
- [x] Feature
- [] Chore
- [] Bug

onefm_developer workflow for git hub

onefm_developer workflow for git hub

## Solution description
onefm_developer workflow for git hub

    - [] Yes
    - [x] No

NA

## Areas affected and ensured
None

No

- [x] Existing Data
- [] New Data

## Was child table created?
    - [] is attachment required?
        did you test attachment
    - [] Yes
    - [x] No

## Is patch required?
- [] Yes
- [x] No
    ## Was the patch test?

  - [x] Chrome
  - [x] Safari
  - [x] Firefox


ks093 (changes requested):
@samdanikouser 
- github.actor is used directly inside the run: block. This is a command injection risk — move it to an env: variable instead.
 
- issue_number from workflow_dispatch is free-text input but never validated. Should check it's actually a number before using it.
 
- The JSON body is constructed via string interpolation. Use jq -n to build it properly — avoids both shell injection and malformed JSON.
 
- Actions are pinned to @v2 tags which are mutable. Pin to full commit SHAs instead (e.g. auth@7c6bc77...). See [GitHub's recommendation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

- No concurrency group defined. If someone labels a bunch of issues at once, all of them will trigger parallel Cloud Run calls. Add a concurrency group per issue number.
 
- Anyone who can add the onefm_developer label can trigger the agent. Consider adding a permission check step (via gh api repos/.../collaborators/.../permission) to ensure the actor actually has write access.

ks093 (approved):

PR #5789, opened 2026-03-18
fix: Item Price duplicate check logic for empty fields

samdanikouser:
### Description
This PR fixes the duplicate check logic for `Item Price` in the `one_fm` app. Previously, the logic omitted empty fields from the `WHERE` clause, causing generic item prices (e.g., without a customer) to conflict with customer-specific prices.

### Changes
- Modified `one_fm/api/doc_methods/item_price.py`:
    - Updated `check_duplicates` to explicitly handle `NULL` or empty values in the database when the current document's field is empty.
    - Added `currency` to the list of fields checked for duplicates.
    - Ensured numeric fields (`packing_unit`, `shift_hours`) check for `0` or `NULL`.

### Verification Results
- Verified that the SQL query now correctly handles empty fields using `(field IS NULL OR field = '')` or `(field IS NULL OR field = 0)`.
- Verified database schema for `tabItem Price` and custom fields.
- Note: `bench migrate` failed due to a pre-existing duplicate in `patches.txt`, which should be addressed separately.

### Review Checklist
- [x] No `frappe.db.commit()` in controllers.
- [x] No N+1 query loops.
- [x] SQL uses parameterized queries.
- [x] Correct file paths and module organization.

### DIFF
--- a/one_fm/api/doc_methods/item_price.py
+++ b/one_fm/api/doc_methods/item_price.py
@@ -37,6 +37,8 @@ def check_duplicates(self):
         else:
             if field in ["packing_unit", "shift_hours"]:
                 conditions += " and ({0} is null or {0} = 0)".format(field)
+            elif field in ["valid_from", "valid_upto"]:
+                conditions += " and {0} is null".format(field)
             else:
                 conditions += " and ({0} is null or {0} = '')".format(field)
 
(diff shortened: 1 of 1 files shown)


NoBoneZ (approved):

NoBoneZ:
@copilot 
Using the commits applied to the check_duplicates method, could you please create a PR targeting version-15? I only want the changes made to that method to be merged into version-15.

PR #6001, opened 2026-04-26
Fix/resignation stabilization and handover

kevinmakora1-crypto:
Is this a Feature, Chore or Bug?
 Feature
 Chore
 Bug
Clearly and concisely describe the feature, chore or bug.
Stabilized the Resignation workflows (Create, Extend, Withdraw) for seamless mobile app integration. Fixed critical bugs causing file attachments to fail/misalign and 417 Expectation Failed workflow state errors. Additionally introduced a new automated Leave Handover routing engine for Offboarding Officer notifications.

Analysis and design (optional)
N/A

Solution description
Workflow Transition Alignment: Corrected the API resignation submission action to Submit to Supervisor to perfectly match the canonical system workflow, permanently resolving 417 Expectation Failed errors generated by the mobile frontend.
JSON Attachment Parsing: Updated one_fm/api/v1/resignation.py to robustly handle stringified JSON payloads from the mobile frontend. Safely decodes base64 attachments and explicitly routes them to the correct resignation_letter field to prevent AttributeErrors.
Leave Handover Routing (Reliever System): Implemented resolve_active_user in utils.py. The system intercepts email dispatch and queries the Leave Application DocType. If the target Offboarding Officer is actively on leave today, it dynamically reroutes the notification to their designated custom_reliever_.
Active Tracker Status Sync: Reconciled a workflow state mismatch where "Withdrawn" documents were not recognized as TERMINAL_STATES by the API, causing the mobile app to block users from initiating new resignations.
Clean PR Sync: Resolved merge conflicts in patches.txt (safely keeping both add_resignation_buttons and fix_dmarc_report_name_column) and cleanly reverted unintended authentication/configuration optimizations to guarantee absolutely zero regressions to existing mobile app login and check-in workflows.
Is there a business logic within a doctype?
 Yes
 No
Output screenshots (optional)
N/A

Areas affected and ensured
Employee Resignation API endpoints (create_resignation, extend_resignation, withdraw_resignation)
Employee Resignation / Extension / Withdrawal core controllers
Offboarding Officer Email Notifications
Active Resignation Mobile Tracker (get_my_active_resignation / get_all_my_resignations)
one_fm/patches.txt
Is there any existing behavior change of other features due to this code change?
Yes. Notification emails natively destined for the "Offboarding Officer" role during resignation workflows will now automatically divert to their assigned reliever if the primary officer has an active, approved Leave Application for the current day.

Did you test with the following dataset?
 Existing Data
 New Data
Was child table created?
 is attachment required? Note: A child table was not created, but the logic mapping attachments (proof documents/letters) into existing fields was completely overhauled and tested to ensure Base64 files save accurately to the backend.
Did you delete custom field?
 Yes
 No If yes, did you write a delete patch?
Is patch required?
 Yes
 No
Was the patch test?
Which browser(s) did you use for testing?
 Chrome
 Safari
 Firefox
 Ionic Mobile App / Capacitor (iOS/Android)


Talleyrand333 (approved):

ks093:
# Code Review: PR #6001 — `fix/mobile-resignation-backend-stabilization`

**Scope:** ~37 files changed | +1,800 / -70 lines
**Reviewed against:** Frappe v15 conventions, `one_fm` coding standards, and security rules

---

> **Verdict: REQUEST CHANGES.** The PR introduces significant resignation workflow features (API layer, new DocTypes, workflow patches, email notifications, recruiter assignment), but contains critical security violations, Frappe anti-patterns, code duplication, and missing test coverage that must be resolved before merge.

---

## 🔴 Critical Issues (Must Fix)

### C1. `ignore_permissions=True` on Every Insert — Security Bypass

**Every single `insert()` call in `resignation.py` uses `ignore_permissions=True`:**

```python
# resignation.py - create_resignation
doc.flags.ignore_validate = True
doc.insert(ignore_permissions=True)  # ← bypasses ALL permission checks

# extend_resignation
ext.insert(ignore_permissions=True)

# withdraw_resignation
withdrawal.insert(ignore_permissions=True)
```

The `verify_employee_authorization()` check only validates the user matches the employee — it doesn't verify the user has DocType-level create permission. An employee with a valid token can create resignation documents even if the Employee role doesn't have create permission on the DocType.

**Fix:** Remove `ignore_permissions=True` and let Frappe's native permission system handle authorization. If the mobile app needs to bypass, implement proper role-based checks first.

### C2. `frappe.get_all()` Used in User-Facing Whitelisted Methods

`frappe.get_all()` **bypasses user permissions**. Every API endpoint uses it:
- `get_supervisor_dropdown()` — returns ALL system users
- `get_my_active_resignation()` — uses `frappe.get_all("Employee Resignation Item")`
- `get_all_my_resignations()` — same
- `extend_resignation()` — same
- `withdraw_resignation()` — same

**Fix:** Replace with `frappe.get_list()` in all whitelisted methods.

### C3. `get_supervisor_dropdown()` — Information Disclosure

```python
@frappe.whitelist()
def get_supervisor_dropdown():
    return frappe.get_all(
        "User",
        filters={"enabled": 1, "user_type": "System User"},
        fields=["name", "full_name"]
    )
```

- No permission check — any authenticated user can call it
- Returns ALL system users, not just supervisors
- No type annotations

### C4. Missing Type Annotations on All 8 Whitelisted Methods

### C5. `frappe.db.commit()` Inside API Endpoints — Transaction Safety

Multiple manual commits. Frappe auto-commits after successful whitelisted method calls. Manual commits risk partial state on exceptions. The `extend_resignation` and `withdraw_resignation` only commit inside `if attachment:` blocks.

### C6. `doc.flags.ignore_validate = True` — Validation Bypass

Skips the controller's `validate()` entirely. The correct fix is to handle attachment before insert or restructure validation, not bypass it.

### C7. Duplicate Validation — `validate()` and `before_save()` Both Check `relieving_date`

Both check the same condition for "Pending Operations Manager" state. Remove from `before_save()`.

### C8. N+1 Query Pattern in All Resignation Lookups

```python
for item in items:
    doc = frappe.get_doc("Employee Resignation", item.parent)  # ← N queries!
```

**Fix:** Use `frappe.get_list()` with fields to fetch all needed data in one query.

---

## 🟡 Significant Issues (Should Fix)

### S1. `mobile_utils.py` — Needs Proof Before Merging (155 lines)

The module claims Frappe v15 doesn't populate `frappe.form_dict` for token auth. **This is not true** when Content-Type is set correctly. Before merging this workaround:

1. **Add a diagnostic endpoint:**
```python
@frappe.whitelist()
def debug_request_params(employee_id: str = None, **kwargs):
    return {
        "form_dict": dict(frappe.form_dict),
        "request_args": dict(frappe.request.args) if hasattr(frappe, "request") else {},
        "content_type": frappe.request.content_type if hasattr(frappe, "request") else None,
        "explicit_employee_id": employee_id,
        "raw_body_preview": frappe.request.get_data(as_text=True)[:500] if hasattr(frappe, "request") else None,
    }
```
2. Call from the mobile app with the same `CapacitorHttp` config
3. Check if Content-Type is `application/json` — Frappe parses this correctly ✅. Missing/wrong Content-Type → empty `form_dict` ❌
4. Fix at the source (mobile `http.service.ts`) if headers are wrong
5. If confirmed as genuine Frappe limitation, file upstream issue and document

**Until proven necessary, do not merge `mobile_utils.py`.** If needed, move to `one_fm/utils.py`.

### S2. Notification Code Duplication

Three identical patterns across `employee_resignation.py`, `employee_resignation_extension.py`, and `employee_resignation_withdrawal.py`. Extract to shared utility.

### S3. `clear_manual_assignments()` — Duplicated, Silently Swallows Exceptions

### S4. Unnecessary `frappe.db.has_column()` Checks

Checking for columns defined in DocType JSON being added **in this same PR**.

### S5. `view_exit_tab()` — 30-Line JS Block Copy-Pasted

Identical in `employee_resignation.js` and `employee_resignation_withdrawal.js`.

### S6. Commented-Out Code

`frm.toggle_display` commented out instead of removed. `employee_resignation_extension.js` contains only Frappe scaffolding boilerplate — delete the file or implement actual logic.

### S7. ~~Patches Use `print()`~~ — Acknowledged as intentional

### S8. Patches Use Raw SQL (`frappe.db.sql`) 

`adjust_ccp_column_widths.py`, `fix_sequence_type_list_view.py`, `hide_ccp_columns.py` all use raw SQL. Use `frappe.db.set_value()` or Property Setters.

### S9. `erf` and `recruiter` in `field_order` Without Field Definitions

`employee_resignation.json` adds these to `field_order` but no corresponding field definitions in the `fields` array. Will cause migration error.

### S10. Workflow Transition Mismatch

API calls `apply_workflow(doc, "Submit to Supervisor")` but the workflow patch defines the first transition as `"Submit to Offboarding Officer"`. These don't match.

---

## 🟢 Minor Issues

| # | Issue |
|---|---|
| M1 | Blank lines added to `__init__.py` with no purpose |
| M2 | Extra newlines in JSON files |
| M3 | `resolve_active_user()` in `api/v1/utils.py` — not API-specific, should be in `one_fm/utils.py` |
| M6 | Redundant `from frappe import _` inside function bodies when already imported at top |
| M7 | `test_employee_resignation_api.py` duplicates test in `test_employee_resignation.py` |
| M8 | `"# Removed demo assignment loop per PR review"` meta-comment in patch |
| M9 | `api/__init__.py` renamed from `templates/pages/__pycache__/` — __pycache__ should never be committed |
| M10 | `notification_log.py` adds broad `except Exception` |
| M12 | Unrelated patches in this PR (CCP columns, sequence type, candidate titles, seed services) — scope creep |

---

## Summary

| Priority | Count |
|:---:|:---:|
| 🔴 Critical | 8 |
| 🟡 Significant | 9 |
| 🟢 Minor | 12 |

PR #6231, opened 2026-06-14
fix(pmr): prevent TimestampMismatchError on Request Change & fix desi…

kevinmakora1-crypto:
Description
This PR resolves two independent issues in the Project Manpower Request (PMR) module and the Designation query:

Fix TimestampMismatchError on Workflow "Request Change" Action

Problem: When triggering the "Request Change" workflow action, the system first executed a separate frappe.call (set_edit_reason) to save the rejection reason. This successfully updated the modified timestamp in the database. When the client subsequently triggered apply_workflow, it sent the old client-side timestamp, causing a TimestampMismatchError and rolling back the workflow transition.
Fix: Updated the whitelisted set_edit_reason API to return the newly updated database modified timestamp, and updated the client-side workflow callback to synchronize frm.doc.modified with the response before continuing.
Fix OperationalError: Unknown column 'disabled' in 'WHERE' in designation query

Problem: The custom designation_query in queries.py filtered on disabled = 0. Since the standard Designation DocType does not have a disabled field in ERPNext/HRMS v15, designation search widgets crashed.
Fix: Removed the invalid disabled filter and added proper Python type annotations to the method.
Verification & Testing
PMR Change Request Transition: Simulated the transition sequence locally; the transition to Draft now succeeds and the rejection reason is saved cleanly.
Designation Search: Verified the custom designation_query successfully returns all active database designations.


one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.js
```diff
@@ -118,13 +118,16 @@ frappe.ui.form.on('Project Manpower Request', {
...
+						method: "one_fm.one_fm.doctype.project_manpower_request.project_manpower_request.set_edit_reason",
 						args: {
 							name: frm.doc.name,
 							reason: values.reason_for_rejection
 						},
 						callback: function(r) {
-							frm.set_value('reason_for_rejection', values.reason_for_rejection);
+							frm.set_value("reason_for_rejection", values.reason_for_rejection);
+							if (r.message) {
+								frm.doc.modified = r.message;
```
ks093: @kevinmakora1-crypto Remove this field setting. This should be updated by frappe automatically.

one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.py
```diff
@@ -263,9 +263,11 @@ def update_select_field_options(self):
 				field.options = "\n".join(opts)
 
 @frappe.whitelist()
-def set_edit_reason(name, reason):
+def set_edit_reason(name: str, reason: str) -> str:
 	doc = frappe.get_doc("Project Manpower Request", name)
 	doc.check_permission("read")
 	frappe.db.set_value("Project Manpower Request", name, "reason_for_rejection", reason)
 	frappe.clear_document_cache("Project Manpower Request", name)
+	return frappe.db.get_value("Project Manpower Request", name, "modified")
```
ks093: @kevinmakora1-crypto Remove the last return line. It is not needed.

ks093 (commented):
@kevinmakora1-crypto Merge conflicts are there. Please resolve.

ks093 (approved):

PR #6247, opened 2026-06-15
fix(pmr): prevent TimestampMismatchError on Request Change

kevinmakora1-crypto:
This hotfix resolves the TimestampMismatchError on the Project Manpower Request (PMR) 'Request Change' workflow action. Replaces the separate set_edit_reason API call with a single-transaction approach using a custom client-side attribute passed in the document payload.

### DIFF
--- a/one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.js
+++ b/one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.js
@@ -102,19 +102,8 @@ frappe.ui.form.on('Project Manpower Request', {
 					fieldtype: 'Small Text',
 					reqd: 1
 				}, (values) => {
-					frappe.call({
-						method: "one_fm.one_fm.doctype.project_manpower_request.project_manpower_request.set_edit_reason",
-						args: {
-							name: frm.doc.name,
-							reason: values.reason_for_rejection
-						},
-						callback: function(r) {
-							resolve();
-						},
-						error: function(err) {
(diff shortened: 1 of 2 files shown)


one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.js
```diff
@@ -102,19 +102,8 @@ frappe.ui.form.on('Project Manpower Request', {
...
-							reason: values.reason_for_rejection
-						},
-						callback: function(r) {
-							resolve();
-						},
-						error: function(err) {
-							reject(err);
-						}
-					});
+					frm.doc._reason_for_rejection = values.reason_for_rejection;
```
ks093, line 105: @kevinmakora1-crypto You can just do frm.doc.reason_for_rejection = values.reason_for_rejection directly here.

one_fm/one_fm/doctype/project_manpower_request/project_manpower_request.py
```diff
@@ -28,6 +28,9 @@ def autoname(self):
 		self.project_request_code = self.name
 
 	def validate(self):
+		if getattr(self, "_reason_for_rejection", None):
```
ks093, line 31: @kevinmakora1-crypto This can be removed after modifying as per the above comment.

ks093 (changes requested):

ks093 (approved):

PR #6505, opened 2026-07-20
fix: resolve patches.txt conflict between test-production and version-15

Talleyrand333:
Resolves the conflict in PR #6504.

**What this does:** Merges version-15 into test-production, resolving the patches.txt conflict:
- Keeps test-production's sprint patches (attendance_correction_fields, leave_application_emergency_contact, etc.)
- Adds version-15's Operations-011 release patches (backfill, normalize, remove_pam_visa, etc.)
- Removes duplicate  entry

**Why:** Once merged, the test-production→version-15 PR (#6504) will have no conflicts.


Talleyrand333:
Replaced with cleaner approach — merge into version-15 instead

PR #6530, opened 2026-07-23
fix(recruitment): draft Job Offer auto-creation shouldn't require candidate KYC fields

kevinmakora1-crypto:
## Summary
- Shortlisting a candidate in the Interview Console silently auto-creates a **draft Job Offer** behind the scenes (`on_update_job_applicant` → `create_draft_job_offer_for_applicant` → `_insert_job_offer_from_applicant`), so it's ready to review/edit before actually being issued.
- `JobOfferOverride.validate()` unconditionally re-validated the linked Job Applicant's full KYC field set (`Passport Number`, `Place of Birth`, `Marital Status`, `Passport Holder of`, `Passport Issued on`, `Passport Expires on`, `Gender`, `Religion`, `Date of Birth`, `Educational Qualification`, `University / School`) on every save — including that silent auto-creation, blocking interviewers from Shortlisting candidates who haven't gone through document collection yet.
- Accepted/Rejected/Hold never trigger this auto-draft-creation path, so they were never actually affected — an earlier commit on this PR incorrectly widened the fix to the Job Applicant's own save (already correctly exempted for Shortlisted) instead of the actual Job Offer-side check. That was reverted.
- **Fix (final form): gate the check on the offer's own `docstatus`** rather than a call-site flag. Skip `validate_mandatory_fields()` while `docstatus == 0` (still a draft — freely reviewable/editable by a recruiter), enforce it once `docstatus != 0` (i.e. once it's actually being submitted/issued to the candidate — `Document._submit()` sets `docstatus = 1` before calling `save()`, so `validate()` always sees the real state).
  - An earlier version of this fix used a `flags.skip_recruitment_mandatory_check` flag set only at the auto-creation call site — that only covered the initial silent save; a recruiter's *next* edit to that same draft (a fresh `frappe.get_doc()` load, no flag) hit the identical wall, contradicting the whole point of a "ready to review/edit" draft. The docstatus-based gate is a property of the document itself, so it doesn't depend on remembering to set a flag at every place that touches a draft.

- **Hardening (independent-review follow-up):** `auto_email_job_offer()` (fired from `on_update()` on every save, including drafts) only gated on an admin-configurable Hiring Settings value (`job_offer_workflow_state`), not on `docstatus` itself. Verified currently inert (that setting is a docstatus=1 state today), but nothing in code prevented it from being pointed at the draft ("Open") state, which would have silently auto-emailed a KYC-incomplete offer letter. Added an explicit `if self.docstatus != 1: return` guard so this can never fire before an offer is actually submitted, regardless of that config.

## Test plan
- [x] Verified via `bench console` with a synthetic applicant missing Passport Number: `_insert_job_offer_from_applicant()` (the Shortlist-triggered auto-draft path) creates the draft without throwing.
- [x] Verified a recruiter's follow-up edit to that same still-draft offer (fresh doc load, docstatus 0) also saves without throwing — the actual gap the flag-based version missed.
- [x] Verified calling `.submit()` on that same draft still correctly throws the mandatory-fields error — KYC completeness is still enforced before an offer is actually issued.
- [x] Cross-checked against real data: found 10 real Shortlisted Bulk Recruitment candidates in the DB with no Job Offer at all (because pre-fix, auto-creation always failed for them, missing KYC data); re-triggering Shortlist on one of them now correctly creates its draft.
- [x] Confirmed `interview_console.py` and `validate_job_applicant()` are unmodified from their original state — no behavior change for Accepted/Rejected/Hold.
- [x] Self-reviewed against the org's `review-erpnext-pr` checklist — no protected files, no permission bypasses, minimal diff.

### DIFF
--- a/one_fm/one_fm/page/interview_console/interview_console.py
+++ b/one_fm/one_fm/page/interview_console/interview_console.py
@@ -210,6 +210,10 @@ def _update_applicant_status(applicant: str, status: str, height: str | None):
 
     if height:
         doc.one_fm_height = float(height)
+    # Interview-stage candidates haven't gone through document collection yet --
+    # the full KYC mandatory-field check (passport, DOB, religion, etc.) belongs
+    # at Job Offer creation, not here.
+    doc.flags.skip_recruitment_mandatory_check = True
     doc.save(ignore_permissions=True)
 
     if status == "Accepted":
@@ -416,6 +420,7 @@ def clear_interview_data(applicant: str):
     if doc.meta.has_field("workflow_state"):
         if doc.workflow_state in ["Accepted", "Job Offer Issued", "Shortlisted", "Hired", "Hold", "Rejected", "Cleared"]:
             doc.workflow_state = ""
+    doc.flags.skip_recruitment_mandatory_check = True
     doc.save(ignore_permissions=True)
 
     return "Success"
--- a/one_fm/one_fm/page/interview_console/interview_console.py
+++ b/one_fm/one_fm/page/interview_console/interview_console.py
@@ -210,10 +210,6 @@ def _update_applicant_status(applicant: str, status: str, height: str | None):
 
     if height:
         doc.one_fm_height = float(height)
-    # Interview-stage candidates haven't gone through document collection yet --
-    # the full KYC mandatory-field check (passport, DOB, religion, etc.) belongs
-    # at Job Offer creation, not here.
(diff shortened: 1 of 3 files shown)


kevinmakora1-crypto:
Closing — rebuilt against staging (not version-15, per team policy) with the equivalent fix applied to staging's actual code, since staging and version-15 have diverged significantly on these files. See #6532.

PR #6557, opened 2026-07-26
fix: repair mangled import in contract_compliance_checker

Talleyrand333:
Fixes a `SyntaxError` that is currently live on `staging`.

`one_fm/operations/doctype/contract_compliance_checker/contract_compliance_checker.py` line 5 reads:

```python
mport refrom datetime import timedelta
```

Two import statements were welded into one invalid line by commit `a0b905fb8`
("fix: conflict resolutions for PR #6469"). Because it is a `SyntaxError`, the
**entire module fails to import** on `staging`.

Restored as two statements — both names are used in this module (`re.search` x2,
`timedelta` x2):

```python
import re
from datetime import timedelta
```

### Note on the other branches
`test-production` and `version-15` are syntactically valid but carry a **latent
`NameError`**: they call `re.search` twice with no `import re` at all. The
Sprint 15th-21st cleanup PR (#6556) already carries this same fix forward, so
merging that PR resolves them.

Found while verifying the Sprint 15th-21st sprint branch; a direct push to
`staging` is blocked by repository rules, hence this PR.

### DIFF
--- a/one_fm/operations/doctype/contract_compliance_checker/contract_compliance_checker.py
+++ b/one_fm/operations/doctype/contract_compliance_checker/contract_compliance_checker.py
@@ -2,7 +2,8 @@
 # For license information, please see license.txt
 import calendar
 import json
-mport refrom datetime import timedelta
+import re
+from datetime import timedelta
 
 import frappe
 from frappe.model.document import Document


Talleyrand333:
Closing as redundant — the fix is already on `staging`.

Commit `e52d84fc8` (the same repair, made on `sprint_15th-21st`) reached `staging` on 2026-07-27 via #6565 `merge/version-15-into-staging`, having travelled sprint branch → `test-production` → `version-15` → back to `staging`.

Verified: this branch's copy of `contract_compliance_checker.py` is byte-identical to `staging`'s (sha256 `f59ff6f402cef97b`), so merging would be a no-op. The `SyntaxError` and the missing `import re` are both resolved on `staging`, `test-production` and `version-15`.

PR #6581, opened 2026-07-29
fix(hr): gate Supervisor/Operations Manager Remarks visibility to their own stage

kevinmakora1-crypto:
## Summary
- `operations_manager_remarks` was gated only on `shift_working`, with no workflow_state check, so it showed as soon as the record existed -- well before the Operations Manager stage was ever reached. Now gated to `Pending Operations Manager`/`Approved`/`Withdrawn`.
- `supervisor_remarks` is now read-only once `workflow_state` moves past `Pending Supervisor`, so the Operations Manager can see the supervisor's remarks but not edit them.
- `operations_manager_remarks` is read-only once past `Pending Operations Manager`, for the same reason -- symmetric with `supervisor_remarks`.

## Test plan
- [ ] At `Pending Supervisor`: confirm `supervisor_remarks` is visible and editable, `operations_manager_remarks` is hidden.
- [ ] At `Pending Operations Manager`: confirm `supervisor_remarks` is visible but read-only, and `operations_manager_remarks` is visible and editable.
- [ ] At `Approved`/`Withdrawn`: confirm both remarks fields are visible and read-only.

### DIFF
--- a/one_fm/one_fm/doctype/employee_resignation/employee_resignation.json
+++ b/one_fm/one_fm/doctype/employee_resignation/employee_resignation.json
@@ -234,7 +234,8 @@
    "fieldtype": "Small Text",
    "label": "Supervisor Remarks",
    "max_height": "4rem",
-   "mandatory_depends_on": "eval:!doc.__islocal && !['Draft', 'Pending Supervisor', 'Pending Relieving Date Correction'].includes(doc.workflow_state)"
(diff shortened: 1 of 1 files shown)


kevinmakora1-crypto:
Consolidating into #6580 per request -- this commit has been cherry-picked onto that branch.

PR #6593, opened 2026-07-30
WI-001768: exempt Day Off OT from the Default Shift Checker

yusuffGoodie:
Implements WI-001768 [Sivaraj] Exemption for Day OFF OT from Default Shift Checker.

## The change

Schedule lines flagged **Day Off OT** are excluded from the counts behind all three categories. Overtime worked on a day off is authorised, so it is not evidence of a mis-allocated shift — counting it flagged relievers for doing exactly what was asked of them, and told the supervisor to "re-assign the excess days" that *were* the approved overtime.

Applied in **both** queries:

- `create_checker`'s `conditions` — decides who gets flagged and against which threshold.
- `get_shift_assignments` — lists the offending dates in the child table.

Both, because a mismatch between them would print a count in the comment that did not match the dates shown beside it.

## Measured impact on the dev dataset

```
Basic + Working schedule lines : 1,997,025
  of which Day Off OT          :    38,204  (1.91%)

relievers whose same-shift count drops:
  HR-EMP-00171   before 1251 -> after 1149   (102 excluded)
  HR-EMP-00266   before  215 -> after  114   (101 excluded)
  HR-EMP-00690   before  965 -> after  879   ( 86 excluded)
  ...

configured thresholds: default 1, day-off reliever 20, weekend reliever 26
```

So this is not a theoretical fix — individual relievers were carrying 80–100 wrongly-counted days against thresholds of 20 and 26.

## What was already in place

Reading the process map against the code, most of it is already implemented and needed no change: the three categories, all three thresholds from ONEFM General Setting, the dynamic weekend-reliever threshold (**27** for a 31-day month, **26** for 30 or fewer — exactly as the note from Sivaraj on the diagram specifies), the `custom_exclude_from_default_shift_checker` project flag, the repeat-count roll-forward from yesterday's record, and the Category B and C comment wording.

The one wording gap: Category A now carries the **"Shift Mismatch:"** label that both the diagram and the AC spell out and the code omitted.

## Two things in the diagram that the ACs do not cover

Flagging rather than silently implementing or ignoring:

1. **Project Manager fallback.** The map shows `Is Site Supervisor Set?` → No → *Assign Project Manager* → notify. Today `create_checker` only sets `site_supervisor`, and the assignment rule keys off `site_supervisor_user`, so a site with no supervisor produces a record nobody is assigned. None of this item's ACs mention it. Should that be added here, or is it a separate item?

2. **`>` vs `>=`.** The ACs say the count must *exceed* the threshold; the code uses `Count(...) >= threshold` and has since before this change. The diagram puts the category definitions in "Details in Notes", which is not visible in the export, so I could not confirm the intent. I have **left the existing comparison untouched** — flipping it would change how many records are raised every night, which is well outside a Day Off OT exemption. Worth a decision.

## Testing

`bench run-tests --module one_fm.operations.doctype.default_shift_checker.test_default_shift_checker` → **11 passed** (the file was previously an empty stub).

Covers the exclusion being present in both queries, `day_off_ot` being a real Employee Schedule field, and — against real Day Off OT rows on the instance — that such a date is not listed and that the count always equals the number of dates listed. Plus the weekend-reliever thresholds (31-day, 30-day, February) and all three category comment strings.

No migration needed. The checker runs nightly from the scheduler, so the next run picks this up.

### DIFF
--- a/one_fm/operations/doctype/default_shift_checker/default_shift_checker.py
+++ b/one_fm/operations/doctype/default_shift_checker/default_shift_checker.py
@@ -124,6 +124,11 @@ def create_checker(start_date, end_date, is_day_off_reliever=False, is_weekend_r
 		(Employee.shift.isnotnull()) &
 		(EmployeeSchedule.employee_availability == "Working") &
 		(EmployeeSchedule.roster_type == "Basic") &
+		# WI-001768: Overtime worked on a day off is authorised, so it is not evidence
+		# of a mis-allocated shift. Those lines are excluded from every category's
+		# count, or a reliever covering approved Day Off OT trips the threshold and is
+		# flagged for doing exactly what was asked of them.
+		(EmployeeSchedule.day_off_ot == 0) &
 		(EmployeeSchedule.date.between(start_date, end_date)) &
 		(Project.custom_exclude_from_default_shift_checker != 1)
 	)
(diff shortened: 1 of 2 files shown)


yusuffGoodie:
Per your decision, the threshold comparison now follows the acceptance criteria: `Count(...) > threshold` rather than `>=`, so a count sitting exactly on the limit is compliant.

This bites hardest on **Category A**, whose configured threshold is **1** — an employee with a single day outside their default shift allocation was previously raised as a shift mismatch and now needs at least two. Category B (20) and Category C (dynamic 26/27) shift by one day each.

Tests updated: **12 passing**, including one asserting the strict comparison so this cannot silently regress to `>=`.

NoBoneZ (approved):

PR #6596, opened 2026-07-30
WI-001790: make the attendance sheet a date-range report

yusuffGoodie:
Implements WI-001790 Monthly Att 1 [Ahmed] Monthly Attendance Sheet Report and Filtering.

## Why this was a rewrite, not a filter change

The report keyed **every cell by day-of-month** (`Extract("day", ...)`, columns named `"1"`..`"31"`, and a `range(1, days_in_month + 1)` loop). A From/To range breaks that outright: `28 Dec` and `28 Jan` collide on the same key.

So cells are now keyed by **ISO date** end to end — the four source queries select the date itself and filter `between(from_date, to_date)`, the columns are named `2026-01-10`, and the per-day loop walks the derived range. `get_date_range()` is the single source of that list, so columns, queries and lookups cannot drift apart.

## Changes

- **From Date / To Date** replace Month + Year, everywhere.
- **New filters**: Employee, Employee Status, Employment Type, Roster Type, Day Off OT, Generate Based On. Roster Type and Day Off OT narrow the source rows; **WI-001791 owns their display semantics** (the four Basic/Overtime rules and the Attendance-Status-vs-Shift-Hours toggle).
- **New columns**: Project, Status, Employment Type, plus Roster Type and Day Off OT, alongside the existing daily detail. Fixed columns kept narrow, since the AC asks the table to fit the screen and every pixel there is taken from the day cells.
- **Generate gate**: the report opens empty and runs only when Generate is clicked. Frappe query reports auto-run on every filter change, and a payroll extract over ~4,000 employees is far too expensive for that. Changing any filter clears the flag again, so what is on screen always belongs to the filters that produced it.
- **Range capped at 62 days**, and a reversed or missing range is refused with a clear message instead of rendering an empty grid.
- **Removed** `get_attendance_years`, dead once the year filter went.

## The print template

Worth calling out, because it was the hidden cost. Its column arithmetic was pinned to a 31-day month:

```
{% var totalColumns = 44; %}
{% var dynamicColumns = totalColumns - dayColumns - fixedColumns; %}
```

A 7-day range inflated the Employee ID/Name columns to absurd widths; a 62-day range drove `dynamicColumns` **negative** and broke every colspan. It now derives its width from the day count, the meta row shows a Period instead of Month/Year, and the data cells index by the new ISO key (`monthDays[i].key`) rather than the day number.

## Testing

`bench run-tests --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet` → **18 passed**.

Covers the range being inclusive, single-day and month-crossing ranges, all four validation refusals plus the maximum allowed range, ISO-keyed and uniquely-named columns, day labels carrying the month, every AC-listed attribute column, the Generate gate (empty result, no day columns, and no premature date validation), the print-template headers, and two checks against real data.

Verified end to end on the dev site — the case the old keying could not represent now works:

```
2026-01-01 .. 2026-01-31  ->  41 cols, 2039 rows   (31 day cols, all populated)
2026-01-10 .. 2026-01-16  ->  17 cols, 1650 rows   (part month)
2025-12-28 .. 2026-01-05  ->  21 cols, 1859 rows   (crosses month boundary)
employee_status = Active  ->  2039 rows becomes 1820
stray keys not in columns: none (in every case)
```

No migration needed; a `bench build`/`clear-cache` picks up the report assets.

## Follow-on

**WI-001791** builds on this: the Attendance Status vs Shift Hours toggle and the four Roster Type / Day Off OT display rules. Note its Logic Rule 4 (Overtime + Day Off OT returns nothing) contradicts the data — Employee Schedule genuinely has `roster_type = "Over-Time"` rows with `day_off_ot = 1` — so that one needs confirming before it is built.

### DIFF
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.html
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.html
@@ -110,21 +110,17 @@
 {% var monthDays = report.additional_details ? report.additional_details.days : [] %}
 {% var statusMap = report.additional_details ? report.additional_details.status_map : [] %}
 
-{% var totalColumns = 44; %}
-{% var fixedColumns = 1 + 2 + 1; %}
 {% var dayColumns = monthDays.length; %}
-{% var dynamicColumns = totalColumns - dayColumns - fixedColumns; %}
+{% var fixedColumns = 1 + 2 + 1; %}
+{% var dynamicColumns = 12; %}
+{% var totalColumns = dayColumns + fixedColumns + dynamicColumns; %}
 
 {% var empIdColSpan = Math.floor(dynamicColumns * 4 / 12); %}
 {% var empNameColSpan = Math.floor(dynamicColumns * 6 / 12); %}
 {% var shiftColSpan = dynamicColumns - empIdColSpan - empNameColSpan; %}
 
-{% var selectedMonth = "" %}
-{% for (var i = 0; i < report.get_filter("month").options.length; i++) { %}
-  {% if (report.get_filter("month").options[i].value == report.get_filter("month").value) { %}
-    {% selectedMonth = report.get_filter("month").options[i].label %}
-  {% } %}
-{% } %}
+{% var periodLabel = frappe.datetime.str_to_user(report.get_filter_value("from_date"))
+     + " - " + frappe.datetime.str_to_user(report.get_filter_value("to_date")); %}
 
 {% function chunk(array, size) {
   var out = [];
@@ -146,7 +142,7 @@
           <td colspan="5" class="logo-cell">
             <img src="https://one-fm.com/files/ONEFM_Identity.png" alt="one-fm" />
           </td>
-          <td colspan="34">{%= __("Monthly Attendance Sheet") %}</td>
+          <td colspan="{{ totalColumns - 10 }}">{%= __("Monthly Attendance Sheet") %}</td>
           <td colspan="5" class="logo-cell">
             {% const clientLogo = report.additional_details.client_logo || "" %}
             {% if clientLogo %}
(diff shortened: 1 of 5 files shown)


yusuffGoodie:
## Two fixes from testing

### 1. Roster Type and Day Off OT were blank on every row

Both were declared as columns, and both were honoured as *filters*, but nothing ever put them on a row — so the filters above worked while the columns beneath stayed empty.

The reason a row could not name them: a row was one employee and one **shift**, and an employee can hold Basic on one day and Over-Time on another within the same shift. There was no single honest value to show.

A row is now one **shift + roster type + Day Off OT flag**. The two fields come out of the Attendance and Employee Schedule queries and join the key the rows are grouped by — which is also the split WI-001791 needs to isolate basic pay from overtime.

Run over your range, 24-01 → 24-02:

```
rows: 2714
  roster_type populated : 2714/2714      Basic 2639, Over-Time 75
  day_off_ot flag set   : 479/2714

  M D Abu Taleb    shift=Day     roster_type='Basic'      day_off_ot=1
  M D Abu Taleb    shift=Night   roster_type='Over-Time'  day_off_ot=0
  M D Abu Taleb    shift=Day     roster_type='Basic'      day_off_ot=0
```

Note the same employee now splits into separate rows where the roster type differs, which is the point. Filters and columns agree:

| Filter | Rows | (roster_type, day_off_ot) present |
|---|---|---|
| none | 2714 | Basic/0: 2160, Basic/1: 479, Over-Time/0: 75 |
| Roster Type = Over-Time | 75 | Over-Time/0: 75 |
| Roster Type = Basic | 2639 | Basic/0: 2160, Basic/1: 479 |
| Day Off OT | 479 | Basic/1: 479 |

### 2. The Generate button duplicated the framework's

Correct — it's gone, along with the hidden `generate` filter and the server-side gate.

It existed to stop a payroll extract over every employee running on each filter change. That is exactly what a **prepared report** does, and the report was already being served as one — which is why you saw *"This report was generated 2 minutes ago… click on Rebuild"* right next to a hand-rolled button doing the same job.

`prepared_report` is now set on the report itself rather than left to Frappe's slow-query threshold, so the deferral is deliberate rather than incidental, and the only controls are the framework's **Generate New Report** / **Rebuild**.

### Tests

```
bench run-tests --app one_fm --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests
Ran 22 tests in 13.503s — OK
```

Five new: both columns are declared, every row names its roster type, Day Off OT is always 0/1, and each filter returns only rows whose column matches. The old gate tests are replaced by ones asserting the report is a prepared report and that no `generate` filter or button remains.

`bench migrate` is needed to pick up `prepared_report` on the Report record.

NoBoneZ (approved):

PR #2026, opened 2023-03-13
hot fix: PIFSS Monthly Deduction - Notify Finance

pjamsheer:
## Is this a Feature, Chore or Bug?
- [x] Bug

## Clearly and concisely describe the feature, chore or bug.
- PIFSS Monthly deduction not allowing to submit
Screenshot 2023-03-13 at 10 54 18 AM

## Solution description
- There were two issues, syntax issue were there, the referenced doc is not created
- Code re write for fetching the doc and check for the existence of referenced doc 

## Areas affected and ensured
- `one_fm/grd/doctype/pifss_monthly_deduction/pifss_monthly_deduction.py`

## Is there any existing behavior change of other features due to this code change?
No

## Did you delete custom field?
- [x] No

## Is patch required?
- [x] No

## Which browser(s) did you use for testing?
- [x] Chrome

PR #6601, opened 2026-07-31
WI-001796: route penalties through the HR, Legal and GM tiers

yusuffGoodie:
## What

Applies the supplied Penalty & Investigation workflow — **10 state rows, 13 transitions, verbatim**.

| From | Action | To | Role |
|---|---|---|---|
| Draft (0) | Submit | Pending HR Administrator (0) | HR User |
| Pending HR Administrator (0) | Submit | **Pending Payroll Officer (1)** | HR User |
| Pending HR Administrator (0) | Submit to Legal | Pending Legal Manager (0) | HR User |
| Pending HR Administrator (0) | Submit to GM | Pending General Manager (0) | HR User |
| Pending HR Administrator (0) | Return To Draft | Draft (0) | HR User |
| Pending HR Administrator (0) | Put On Hold | On Hold (0) | HR User |
| Pending Legal Manager (0) | Submit to GM | Pending General Manager (0) | Legal Manager |
| Pending General Manager (0) | Submit | Pending Payroll Officer (1) | General Manager |
| Pending General Manager (0) | Reject | Rejected (0) | General Manager |
| Pending Payroll Officer (1) | Submit | **Completed (1)** | Payroll Operator |
| Pending Payroll Officer (1) | Cancel | **Cancelled (2)** | HR User |
| On Hold (0) | Approve | Pending Payroll Officer (1) | HR User |
| On Hold (0) | Reject | Rejected (0) | HR User |

Workflows are only created by the installer, so an existing site gets it through a patch. All 308 records sit in Draft, which the definition still carries; the patch reports anything found in a retired state rather than leaving it unusable.

## Two things about this definition worth calling out

**Pending Payroll Officer carries docstatus 1.** That settles a problem I had worked around in the previous revision of this branch: with payroll holding a *submitted* document, `Cancel → Cancelled (2)` is legal, whereas a docstatus-0 payroll state forces the cancelled state to 0 too — Frappe refuses 0 → 2 with *"Cannot cancel before submitting"*. It also means the penalty locks when HR submits it, not only once payroll finishes. Every docstatus pairing here is legal, checked against the same rules `Workflow.validate_docstatus()` applies.

The offence count is unaffected: WI-001794 requires **both** docstatus 1 and an approved state, and Pending Payroll Officer is not one, so a penalty merely awaiting payroll does not count as a prior offence. A test pins that.

**No transition carries a condition**, so the salary tier this story is named for is not evaluated by the workflow. Draft has a single unconditional Submit to the HR Administrator, and the **>350 KD route is reached by an HR User pressing "Submit to Legal"** — i.e. the routing is manual. The criteria describe it as automatic ("*Given basic salary is more than 350 KD ... the HR Administrator flow terminates, and the task is routed directly to the Legal Manager*"). If it should be automatic, it is a condition on these transitions plus a source for the threshold; `Employee.one_fm_basic_salary` ("Base Salary") holds it, workflow conditions are handed `frappe.db.get_value` explicitly, and I verified the expression evaluates under RestrictedPython (140 KD → HR, 1000 KD → Legal, unset → HR). Say the word and it goes back in. A test records that it is manual today so nobody has to rediscover it.

## Still blocked on fields that do not exist

The criteria added at 12:13 gate the routing on **`employee_response`** (Accept / Refuse *with `refusal_proof`* / Not Return from Vacation / Request Investigation). Neither field is on the doctype, nor in `WI-001796-doctype.json`, whose nearest is `issuance_status` (*Accepted / Refused / Request for Investigation by Employee*) — missing "Not Return from Vacation", no proof attachment. Not invented.

Also unresolved: the criteria describe no **Pending Employee Response** state, but the live workflow has one, the process map has one, and both this controller's `validate_workflow_transition` and the form's `toggle_visibility` key on it and on *Employee Rejected*. Those guards are unreachable under this definition; left in place pending a decision.

## Testing

`one_fm/tests/test_penalty_workflow.py` transcribes the supplied definition as a table and asserts the applied workflow matches it **in both directions** — including the two `Cancelled` rows, which save fine (no unique constraint on `Workflow Document State.state`, and `get_state()` reads the first match). Plus: one active workflow, the applied workflow matches the shipped file field for field, each state's docstatus, no transition breaking Frappe's docstatus rules, nothing leaving Completed, self-approval and email on every row, every role exists (Payroll Operator, since there is no "Payroll Officer" role), no waiting state is a dead end, no record stranded, and the offence-count interaction above.

```
bench migrate
bench run-tests --app one_fm --module one_fm.tests.test_penalty_workflow --skip-before-tests
```

> Stacked on **#6600** (WI-001795). Merge order: 6594 → 6600 → this → 6602.


yusuffGoodie:
## Added: Employee Response and Refusal Proof

The two fields the criteria route on now exist on the doctype, in the section that already holds the employee's remark — the answer, its proof, then the rejection remark.

| Field | Type | Behaviour |
|---|---|---|
| `employee_response` | Select — *Accept*, *Refuse*, *Not Return from Vacation*, *Request Investigation* | not mandatory, no default: a penalty starts with no response recorded |
| `refusal_proof` | Attach | shown **and** mandatory only when the response is *Refuse* |

Worded from the criteria rather than from `WI-001796-doctype.json`'s `issuance_status`, which offers *Accepted / Refused / Request for Investigation by Employee* — a different tense, and no equivalent of *Not Return from Vacation*.

The criteria say "*Refuse (with valid refusal_proof attached)*", so the requirement belongs to that one answer; making the field unconditionally mandatory would block the other three. Both conditions are declarative (`depends_on` / `mandatory_depends_on`) rather than scripted, so they hold wherever the document is opened — the same choice WI-001795 made for its read-only rules.

Verified against each answer:

```
response=(none)                     proof mandatory: False
response=Accept                     proof mandatory: False
response=Refuse                     proof mandatory: True
response=Not Return from Vacation   proof mandatory: False
response=Request Investigation      proof mandatory: False
```

### The routing these fields exist for is still not implemented

The supplied workflow carries no transition conditions, so as things stand:

- a penalty answered **Accept / Refuse / Not Return from Vacation** can still be sent to Legal or the GM, where the criteria say it "*can only be submitted directly to Payroll*";
- **Request Investigation** is not routed to HR or Legal by salary tier — an HR User chooses the action by hand.

Two ways to close it, both small now that the fields exist: conditions on the existing transitions (`doc.employee_response`, plus `frappe.db.get_value` for the salary, which workflow conditions are given), or a guard in `validate_workflow_transition`, which already exists in this controller for the older states. Tell me which and I'll add it — I've left the workflow exactly as supplied.

Tests: `one_fm/tests/test_penalty_employee_response.py` — the options are exactly the four the criteria name, no default and not mandatory, the proof is an Attach required only on Refuse and never unconditionally, its condition reads the response field, and the two sit before the rejection remark.

```
bench migrate
bench run-tests --app one_fm --module one_fm.tests.test_penalty_employee_response --skip-before-tests
```

yusuffGoodie:
## Now routing on Issuance Status

`employee_response` is gone. The answer is recorded in **`issuance_status`** — the field `WI-001796-doctype.json` already defines — with the criteria's fourth answer added:

```
Accepted
Refused
Not Return from Vacation                 <- added
Request for Investigation by Employee
```

Ordered so the three that go straight to payroll sit together, ahead of the one that opens an investigation — the same grouping the conditions use. `refusal_proof` now depends on **"Refused"** and stays mandatory for that answer alone.

## The conditions

Added to the three routes out of Pending HR Administrator; nothing else about the supplied definition changed — same 10 states, same 13 transitions, same docstatus values.

| Action | Condition |
|---|---|
| Submit → Payroll | `doc.issuance_status in ("Accepted", "Refused", "Not Return from Vacation")` |
| Submit to Legal | `doc.issuance_status == "Request for Investigation by Employee" or (frappe.db.get_value("Employee", doc.employee, "one_fm_basic_salary") or 0) > 350` |
| Submit to GM | `doc.issuance_status not in ("Accepted", "Refused", "Not Return from Vacation")` |

**Put On Hold** and **Return To Draft** stay unconditional — nothing in the criteria gates them, and gating them could strand a penalty. The salary is read with `frappe.db.get_value`, which workflow conditions are handed explicitly, so no field is copied onto the penalty.

## What HR actually sees

Evaluated off the shipped files, against real employees on each side of the threshold:

| Issuance Status | Tier | Available |
|---|---|---|
| *(none)* | ≤350 | Submit to GM, Put On Hold, Return To Draft |
| *(none)* | >350 | Submit to Legal, Submit to GM, Put On Hold, Return To Draft |
| Accepted | ≤350 | **Submit**, Put On Hold, Return To Draft |
| Accepted | >350 | **Submit**, **Submit to Legal**, Put On Hold, Return To Draft |
| Refused | ≤350 | Submit, Put On Hold, Return To Draft |
| Refused | >350 | Submit, Submit to Legal, Put On Hold, Return To Draft |
| Not Return from Vacation | ≤350 | Submit, Put On Hold, Return To Draft |
| Not Return from Vacation | >350 | Submit, Submit to Legal, Put On Hold, Return To Draft |
| Request for Investigation by Employee | ≤350 | Submit to Legal, Submit to GM, Put On Hold, Return To Draft |
| Request for Investigation by Employee | >350 | Submit to Legal, Submit to GM, Put On Hold, Return To Draft |

Two things to read before this merges:

1. **With no answer recorded**, payroll is closed but the GM route stays open, so a penalty is never stuck. The criteria only speak to the cases where an answer is set, so I left the GM route open rather than trapping an unanswered penalty.
2. **An employee over 350 KD who accepted** is offered both *Submit* and *Submit to Legal*. That is the two criteria overlapping — one says the three answers may only go to payroll, the other says the higher tier is routed to Legal — so HR chooses. If Legal should win outright for >350, the payroll condition needs `and salary <= 350` adding.

Still not automatic: a **>350 penalty starts at Pending HR Administrator**, because Draft has only one Submit transition in the supplied definition and adding a second (Draft → Legal, conditioned on salary) would be a change to the definition rather than a condition on it. Say the word if you want it.

## Testing

```
bench migrate
bench run-tests --app one_fm --module one_fm.tests.test_penalty_workflow --skip-before-tests
bench run-tests --app one_fm --module one_fm.tests.test_penalty_issuance_status --skip-before-tests
```

The workflow tests now cover the routing: the three routes carry conditions and the other two do not, an answered penalty reaches payroll and cannot be escalated, a request for investigation cannot go straight to payroll, the higher tier can always reach Legal, an unanswered penalty is never stuck, every answer leaves at least one way forward, and the conditions read option values that actually exist on the field.

yusuffGoodie:
## Aligned with the BA site's doctype layout

The section order, field order and field set now match the export in `WI-001796-doctype.json`, verified field-for-field against it.

### Renames — not drops

| Was | Now | Production data |
|---|---|---|
| `location` | `operations_site` | **all 325 penalties** |
| `deduction_type` | `action_type` | 0 |
| `issuance_status` | `employee_response` | 0 |
| `employee_rejection_remarks` | `employee_remarks` | 0 |
| `legal_findings` | `legal_department_remarks` | 0 |
| — | `created_by` (new) | backfilled from `owner` |

Dropped, all empty on production and referenced by nothing: `reported_by`, `investigation_requiref`, `description`, `evidence`, `witness`.

`rename_penalty_fields_to_ba_layout` runs **post_model_sync** — the only order that works. `rename_field` copies one column into another, so both must exist, and Frappe never drops the column of a removed field (there is no `DROP COLUMN` anywhere in its schema layer), so after the sync the old column is still there holding its data.

### Six places where the export would have undone an implemented criterion

The export predates WI-001794/95/96, so where it conflicts, the implemented setting wins:

| Field | Export says | Kept as | Why |
|---|---|---|---|
| `applied_level` | `1..5` | `1st..5th` | the Penalty Code matrix is keyed on ordinals — bare numbers match no row, so **every sanction lookup would silently return nothing** |
| `action_type` | `Data`, read-only | `Select` + `read_only_depends_on` | WI-001795 needs a dropdown; a Data field ignores its options |
| `penalty_category` | no condition | `read_only_depends_on` | derived when a code is applied, entered by hand when there is none |
| `salary_deduction_amount` | no conditions | `depends_on` + `read_only_depends_on` | WI-001794 requires it read-only; WI-001795 shows it only for a manual Uniform/Damage charge |
| `employee_response` | 3 answers | 4 — adds *Not Return from Vacation* | the criteria use it |
| `refusal_proof` | dropped | kept, keyed on `Refused` | the criteria require proof for a refusal |
| `operations_site` | `fetch_from: employee.site` | no fetch | WI-000513 and Penalty Issuance each set the site from **their own** source; a fetch would overwrite both |

Also not applied: `legal_investigation_report.mandatory_depends_on = eval:doc.investigation_requiref`, which points at a field the same export deletes.

### Everything that moved with the renames

The controller, the form script, the workflow's three routing conditions (`doc.issuance_status` → `doc.employee_response`), **`overrides/shift_request.py`** (WI-000513's retroactive day-off penalty) and **`penalty_issuance.py`** — both of which set the site explicitly — and the tests. `test_penalty_issuance_status.py` is back to `test_penalty_employee_response.py`.

### Verified

```
layout matches the export field for field ......... yes
applied_level ordinals vs matrix .................. 1st..5th, all Penalty Level rows resolve
workflow conditions on employee_response .......... evaluate, no reference to issuance_status
no code path names a removed field ................ clean
```

`bench migrate` is required before the tests will pass — the meta they read comes from the database.

> 6594 and 6600 are untouched and remain self-consistent: they still carry `deduction_type` in both schema and code, so each stands on its own. The rename lands here and flows to 6602. Merge order unchanged: 6594 → 6600 → this → 6602.

yusuffGoodie:
## Fix: `AttributeError: no attribute 'issuance_status'` on form load

The rename reached the doctype but not the workflow. `update_penalty_and_investigation_workflow` had already run under its `#2026-07-30` tag, so `bench migrate` skipped it and the three conditions **in the database** still read `doc.issuance_status` — a field the same migrate had just renamed away.

Patch Log on the test site confirms it:

```
one_fm.patches.v15_0.rename_penalty_fields_to_ba_layout #2026-08-02        <- ran
one_fm.patches.v15_0.update_penalty_and_investigation_workflow #2026-07-30 <- ran on 30 Jul, skipped now
```

The tag is bumped so it re-applies, and it now runs **after** the rename:

```
rename_penalty_fields_to_ba_layout #2026-08-02
update_penalty_and_investigation_workflow #2026-08-02 rerun for employee_response
add_penalty_and_investigation_assignment_rules #2026-08-01 rerun for payroll officer
```

### Why it threw instead of quietly misrouting

`one_fm/overrides/workflow.py` overrides `get_transitions` and evaluates conditions with the **Document itself** as the eval locals — `dict(doc=doc)` — where Frappe core passes `doc.as_dict()`. A `_dict` returns `None` for a missing key; a Document raises. So on this app a condition naming a dropped field is a hard error the moment the form loads. Better than the alternative, and worth remembering when writing conditions here.

### Two tests, because the obvious one doesn't work

`condition` was missing from the fields the drift test compares, so a fixture the database never received still passed. Included now.

More subtly, a condition **cannot** be exercised against a record loaded from the database. Frappe never drops the column of a removed field, so a stored row still carries `issuance_status` as an attribute and the stale condition evaluates happily — which is exactly why this got past me. The form posts only the fields the current meta knows about, and `get_transitions` rebuilds the Document from that payload, which has no orphan columns. The new test rebuilds it the same way:

```
stale   (doc.issuance_status)     guard FAILS -> AttributeError    <- reproduces the report
shipped (doc.employee_response)   guard PASSES
fields named by conditions that do not exist: none
```

After `git pull && bench migrate`, opening PEN-08-2026-0009 should work. If the Patch Log ever shows the workflow patch as applied while the fixture has moved on, that is this same bug — the tag needs bumping.

NoBoneZ (approved):

PR #6602, opened 2026-07-31
WI-001798: put every waiting penalty on someone's desk

yusuffGoodie:
## What

The three assignment rules from `WI-001798.json`, one per waiting state of the WI-001796 workflow, shipped **verbatim** as app fixtures in `one_fm/custom/assignment_rule/` and applied to existing sites by a patch (the installer is the only other thing that creates assignment rules).

| Rule | Assigns on | Closes when |
|---|---|---|
| Penalty and Investigation-HR Administrator | `Pending HR Administrator` | state moves on |
| Penalty and Investigation-Legal Manager | `Pending Legal Manager` | state moves on |
| Penalty and Investigation-General Manager | `Pending General Manager` | state moves on |

All three: `rule: "Based on Field"`, `field: "owner"`, priority 0, all seven days, `is_assignment_rule_with_workflow` exactly as each rule declares it (1, 1, 0).

## Correction (force-pushed)

The first version of this branch changed the selection to `Based on Process Task` and created three Process Tasks naming specific employees. Neither is in the work item. Both are gone, along with the Process Task lookup in the patch.

## One thing to decide, not changed here

The conditions read `doc.workflow_state == "..."`. Frappe evaluates them with the document itself as the eval locals:

```python
# frappe/automation/doctype/assignment_rule/assignment_rule.py
frappe.safe_eval(self.get(fieldname), None, doc)
```

`doc` is therefore not a name in scope, so the expression raises `NameError`, which `AssignmentRule.safe_eval` catches and turns into `False`:

```
>>> eval("doc.workflow_state == 'x'", {}, {"workflow_state": "x"})
NameError: name 'doc' is not defined
>>> eval("workflow_state == 'x'", {}, {"workflow_state": "x"})
True
```

Left as supplied — the rules will not assign anyone until the `doc.` prefix comes off. No other assignment rule in the app uses it (114 fixtures, 0 with the prefix). Say the word and I will drop it.

## Testing

`one_fm/tests/test_penalty_assignment_rules.py` — each saved rule is compared against the supplied definition field for field (document_type, priority, disabled, description, is_assignment_rule_with_workflow, assign_condition, close_condition, rule, field), so a future edit to the spec has to be deliberate. Plus: every tier has a rule, the field they assign on exists, they are enabled and cover all seven days, every state they wait on is a real workflow state, no `Pending` state is left unassigned, and the notification only renders fields the doctype has.

```
bench migrate
bench run-tests --app one_fm --module one_fm.tests.test_penalty_assignment_rules --skip-before-tests
```

> Stacked on **#6601** (WI-001796). Merge order: 6594 → 6600 → 6601 → this.


yusuffGoodie:
## Fixed: "Auto assignment failed: name 'doc' is not defined"

Reported from the UI when saving a penalty — the message appeared three times, once per rule.

### Cause

That text is Frappe catching an exception from an assignment rule condition:

```python
# frappe/automation/doctype/assignment_rule/assignment_rule.py
try:
    if self.get(fieldname):
        return frappe.safe_eval(self.get(fieldname), None, doc)
except Exception as e:
    frappe.msgprint(_("Auto assignment failed: {0}").format(str(e)), indicator="orange")
```

The condition is evaluated with **the document itself as the eval locals**, so `workflow_state` resolves and `doc.workflow_state` is a `NameError`. Reproduced against a real penalty:

```
assignment rule  doc.workflow_state == "Pending HR Administrator"  -> NameError: name 'doc' is not defined
assignment rule  workflow_state == "Pending HR Administrator"      -> False
workflow         Submit to GM  -> True
workflow         Submit        -> False
workflow         Submit to Legal -> False
```

**Not the WI-001796 workflow conditions** — those evaluate cleanly, as the last three lines show. A workflow transition condition *is* handed `{"doc": ...}` (`frappe.model.workflow.get_workflow_safe_globals`) and does want the prefix; an assignment rule is not. The two layers differ, and these rules were on the wrong side of it. This is the trap called out in the PR description when the rules were shipped as supplied.

The effect was silent: the penalty saved, but nobody was ever assigned.

### Fix

The `doc.` prefix is dropped from `assign_condition` and `close_condition` in the three fixtures. Nothing else in the app carries it — these were 3 of 117 — and the six affected rows on the site are the same three rules.

The patch entry is bumped to `#2026-08-01 rerun` so `bench migrate` rewrites the rows already created; `create_assignment_rule` updates an existing rule rather than skipping it.

Verified against every state of the workflow:

| Rule | assigns on | closes on |
|---|---|---|
| HR Administrator | Pending HR Administrator | every other state |
| Legal Manager | Pending Legal Manager | every other state |
| General Manager | Pending General Manager | every other state |

### Test

`TestConditionsActuallyEvaluate` now runs each condition the way the framework does — `frappe.safe_eval(condition, None, doc)` — across every workflow state plus `None` and `""`, and fails on a raise rather than on a string comparison. That is what would have caught this: the previous test only checked the text of the condition.

```
bench migrate
bench run-tests --app one_fm --module one_fm.tests.test_penalty_assignment_rules --skip-before-tests
```

NoBoneZ (approved):

PR #6605, opened 2026-07-31
WI-001791: give the attendance sheet its shift-hours view and roster rules

yusuffGoodie:
## What

WI-001790 added the **Roster Type**, **Day Off OT** and **Generate Based On** filters to the Operations Monthly Attendance Sheet. This is the logic behind them.

## Shift Hours

The day cells now carry the shift's **scheduled** duration, from `Operations Shift.duration` — which the report was already joining — formatted as the criteria write it: `12`, not `12.0`.

Deliberately **not** `Attendance.working_hours`: that is what was clocked (the AC's 11.5), and the criteria ask for what was rostered. Consequences, all covered by tests:

- A day the shift was rostered but not worked still shows its duration — scheduled is scheduled.
- A day with no shift at all (a day off) stays blank; there are no scheduled hours to show.
- Working Days and Days Off keep coming off the status, so switching mode cannot move the totals.
- The status legend is replaced by one describing what the cells hold, and the day-cell colouring is skipped for values the status map does not know.

## The four Logic Rules

Applied in one place shared by the Attendance and Employee Schedule queries, so both answer a given filter pair identically:

| Roster Type | Day Off OT | Condition | Rows (2026-07-01..07) |
|---|---|---|---|
| Basic | unchecked | `roster_type='Basic' AND day_off_ot<>1` | 1719 |
| Basic | checked | `roster_type='Basic' AND day_off_ot=1` | 51 |
| Over-Time | unchecked | `roster_type='Over-Time'` | 62 |
| Over-Time | checked | refused before any query | empty + message |

**Rule 1 was the only one actually wrong**: Basic on its own included the day's day-off OT rows, and its edge case asks for them to be hidden. Rule 3 constrains only the roster type, which already hides both basic cases — its edge case names nothing else.

Rule 4 is answered before any query runs, with the columns still built so the table renders an empty state:

> **Overtime** with **Day Off OT** is not a valid combination. Clear one of them to generate the report.

## Worth flagging

**365 Employee Schedule rows genuinely carry `roster_type='Over-Time'` with `day_off_ot=1`**, so Rule 4's combination is not impossible data. They stay reachable through *Overtime with Day Off OT unchecked*, which Rule 3 does not exclude — if you would rather that combination were also excluded, say so and it is a one-line change.

Two things I did not touch: the **Roster Type** and **Day Off OT** columns on each row are declared but never populated (from WI-001790), and the **print template's** legend still lists status abbreviations in Shift Hours mode. Neither is in the criteria.

## Testing

```
bench run-tests --app one_fm --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests
Ran 36 tests in 8.638s — OK
```

20 new cases: each rule's SQL conditions, Rule 4's empty state, a blank roster type left unconstrained, the scheduled duration in the cells including on an absent day, a day off staying blank, Attendance Status mode unchanged, the totals agreeing across modes, the legend, hour formatting, and the hours-map merge precedence.

Also verified end to end against the real data — the row counts in the table above, and the cells rendering `12` where Attendance Status renders `P`.

> Stacked on **#6596** (WI-001790). Retarget to `staging` once that lands.

### DIFF
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.js
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.js
@@ -107,7 +107,11 @@ frappe.query_reports["Operations Monthly Attendance Sheet"] = {
 	formatter: function (value, row, column, data, default_formatter) {
 		value = default_formatter(value, row, column, data);
 		if (column.colIndex < FIXED_COLUMNS) return value;
-		return `<span style='color:${status_color_map[value]}'>${value}</span>`;
+		// Under "Shift Hours" the day cells hold a duration, which has no status colour
+		// (WI-001791) - colour only what the map knows.
+		const color = status_color_map[value];
+		if (!color) return value;
+		return `<span style='color:${color}'>${value}</span>`;
 	},
 	onload: function (report) {
 		report.page.add_inner_button(__("Generate"), () => {
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
@@ -3,7 +3,7 @@
 
 import frappe
 from frappe import _
-from frappe.utils import add_days, cint, cstr, date_diff, getdate, nowdate
+from frappe.utils import add_days, cint, cstr, date_diff, flt, getdate, nowdate
 
 status_map = {
 	"Present": "P",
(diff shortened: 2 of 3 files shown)


yusuffGoodie:
## Shift Hours restored, and the filters no longer round-trip

### Why Shift Hours showed statuses

Two changes collided. WI-001790's column fix made a row **one shift + one roster type + one Day Off OT flag**, but this branch's hours map was still keyed by **shift alone** — so the lookup missed and every cell fell back to the status.

The hours map is now keyed by the same group as the row it belongs to, which also stops a shift split across roster types from sharing one set of hours. Verified over 24-01 → 24-02:

```
Attendance Status   rows=2714   cells: P 8056, DO 917, CDO 214, A 44, OL 13
Shift Hours         rows=2714   cells: 12 6414, 8 1338, 13 138, 18 63, 9 62, 12.5 54
```

> Worth knowing for the branch order: Shift Hours lives here, not on #6596. If you were testing on `WI-001790` you would not have seen it at all — that branch has the filter but not the logic behind it.

### The filters were queuing a background run each time

Frappe calls a filter's `on_change` **in place of** refreshing the report, so these now narrow the rows already on screen with no server call:

**In page** — Employee, Employee Status, Employment Type, Roster Type, Day Off OT
**Server** — From/To Date, Include Future Attendance, Generate Based On

Those three stay on the server because each changes what has to be *fetched*, not which rows to keep.

**Project stays on the server too**, and that one wasn't obvious: it filters the Attendance rows as well as the employees, so an employee with days booked to another project comes out differently in page — 51 rows against the server's 44 when I tried it. Left where it works.

The in-page predicate repeats the roster-type rules rather than approximating them, including Rule 1's exclusion of day-off-OT rows from plain Basic and Rule 4's empty result. Checked against the server across ten combinations:

```
{}                                        server=2714  in-page=2714  OK
{'roster_type': 'Basic'}                  server=2160  in-page=2160  OK
{'roster_type': 'Over-Time'}              server=  75  in-page=  75  OK
{'day_off_ot': 1}                         server= 479  in-page= 479  OK
{'roster_type':'Basic','day_off_ot':1}    server= 479  in-page= 479  OK
{'roster_type':'Over-Time','day_off_ot':1} server=  0  in-page=   0  OK   <- Rule 4
{'employee_status': 'Active'}             server=2436  in-page=2436  OK
{'employment_type': 'Full-time'}          server=2623  in-page=2623  OK
{'employee': 'HR-EMP-00006'}              server=   3  in-page=   3  OK
{'employee_status':'Active','roster_type':'Basic'} server=1931 in-page=1931 OK
```

Rows now carry the Employee docname in a hidden column — the Employee filter returns a docname while the visible Employee ID is the badge number, so matching on that would never hit.

### Tests

```
bench run-tests --app one_fm --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests
Ran 47 tests in 17.782s — OK
```

Merge order: **#6596 → #6605**.

NoBoneZ (approved):

PR #6616, opened 2026-08-02
WI-001808: render the PDF logo from disk, and let a cell follow its own script

yusuffGoodie:
Two fixes from testing WI-001808. #6583 is already merged, so these come as a follow-up off `staging`.

## 1. "PDF generation failed because of broken image links"

The logo was fetched **over HTTP**. Frappe turns a relative `src` into an absolute URL with `frappe.utils.get_url()`, which on this site is:

```
frappe.utils.get_url() = http://one_fm.15      <- not a resolvable host
```

wkhtmltopdf then exits with `UnknownContentError`, and Frappe converts that into *"PDF generation failed because of broken image links"* — so **that contract's PDF never makes it into the ZIP**. It doesn't happen on staging because there the site name is a real domain.

The logo is now read off disk and inlined as a data URI, which removes the network from the path entirely. That matters doubly because the ZIP is built in a background job.

Verified against the flags Frappe actually runs wkhtmltopdf with:

```
wkhtmltopdf --disable-local-file-access ...
  http URL   -> 15,214 bytes   Warning: Failed to load http://one_fm.15/files/onefmlogo.png (ignore)
  data URI   -> 21,129 bytes   logo embedded
```

Note the first line: **even where the old path did produce a PDF, it rendered without the logo.** So this also puts the logo back on documents that appeared to work.

## 2. RTL alignment in the Summary and Non-Manpower tables

The tables are `direction: rtl`, but every value inside them was wrapped in `.en`, which forces `direction: ltr`, and three columns pinned `text-align: left`. An Arabic value would therefore have been laid out and aligned **left** inside an RTL table — against the criterion that "all table layouts, columns, and text alignment must strictly adhere to a Right-to-Left (RTL) centric design".

Values now carry `dir="auto"` with `unicode-bidi: plaintext` and `text-align: start`, so each cell takes its direction from its own first strong character: Arabic reads right-to-left, Latin left-to-right, in a table that stays RTL.

Nothing changes for today's data — `item_type`, `contract_item_category` and the staff breakdown are entirely Latin right now (`Janitor`, `Janitorial Services`, `- 2 Staff worked 20 days: 40 Days`). The criterion is about what happens when they are not.

The boilerplate paragraph keeps `.en`: an LTR island inside an Arabic sentence is exactly what a date or a company name wants there.

## Testing

`bench migrate` is required — Print Format is a database record, so the template change only reaches the site through a sync.

Then generate a POW PDF: the header logo should appear, and no *broken image links* entry should land in the Error Log.


yusuffGoodie:
## Added: an hourly Sale Item's day cells now read in hours

Where the summary counts a Sale Item in hours, the attendance sheet behind it now does too.

- A day with a worked status shows the **hours recorded against that Attendance**.
- The row's **Total Hours is their sum**, not days × shift length — so the cells add up to the total printed beside them.
- A day that was **not** worked keeps its abbreviation. `0` against an absence reads as a figure rather than an explanation.
- Where the source carries no hours at all — an Attendance Amendment holds statuses only — it falls back to the shift length, which is the fallback the summary already applies, so the two still agree.

Verified on real contracts:

```
Al Babtain
  SER-SEC-000136-NKW-M-30DY-12HR   rate=Hourly    basis=Shift Hours
     cells = 11.51, 9.88, 11.95, 12, 11.92, 9.13, 12.23, 12 ...   total_hours = 338.85
  SER-FMG-000294-NKW-M-22DY-9HR    rate=Monthly   basis=Attendance Day
     cells = P, P, P, P, P, P, P, P                                total_hours = 189
```

## Which items switch, and why it is decided this way

By `_basis_for_rate_type` — **the same function page 1 uses** — not a second reading of the Rate Type. Hourly lands on Shift Hours and reports in hours; Daily and Monthly stay on days and keep their statuses. Deciding it twice is how the two pages would eventually come to disagree, so there is one gate and a test that pins it.

## ⚠️ Your test contract will not show hours — its items are Monthly

Worth knowing before you retest. On `Al - Futtaim … Majed Al-Futtaim`:

```
SER-FMG-000294-NKW-M-22DY-8HR    rate_type = Monthly   -> Attendance Day
SER-SEC-000136-NKW-M-30DY-12HR   rate_type = Monthly   -> Attendance Day
```

Both Contract Items are **Monthly**, so under WI-001700's rule ("Daily and Monthly are counted in present days") they report in days, and the sheet keeps its statuses.

That leaves a separate question about **page 1 of that document**: its stored rows read `54.31 hrs` and `1908.90 hrs`, i.e. hours, because the document carries `generation_basis = Shift Hours` from when it was generated. Regenerated today it would report days, since the Rate Type now takes precedence. So that POW is stale rather than wrong — but if you expect that contract to bill hourly, the Contract Item Rate Types are what need looking at, not the report.

To see the hours view, use a contract with an **Hourly** item — `Al Babtain` above is one on this site.

## Testing

```
bench run-tests --app one_fm --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests
Ran 63 tests in 5.704s — OK
```

Seven new: hours on a worked day, an unworked day keeping its abbreviation, the shift-length fallback, half days counting as worked, empty days staying empty, the worked-status set being derived rather than restated, and the gate calling page 1's own basis function.

yusuffGoodie:
## Why neither fix appeared after `bench migrate`

**The print formats never reached the database.** Frappe's importer skips a file whose `modified` timestamp is not newer than the record's:

```python
# frappe/modules/import_file.py
is_db_timestamp_latest = db_modified_timestamp and (
    get_datetime(doc.get("modified")) <= get_datetime(db_modified_timestamp)
)
```

Both file and record sat at `2026-07-29 22:10:00` — equal, so migrate read the files and did nothing. The logo fix and the RTL work were in the repository and absent from the site at the same time, which is why *both* symptoms survived the migrate. Both stamps now move to `2026-08-03`, and I verified the import lands:

```
Proof of Work Letter               pow_logo_src in DB: True   dir="auto" in DB: True
Proof of Work Attendance Report    pow_logo_src in DB: True
```

## The ZIP failure is environmental, not this template

With the format finally imported, the PDF *still* failed — so I tested a control: a plain **Employee / Standard** print format on the same site.

```
Employee, Standard, as_pdf   ->  ALSO FAILS: "broken image links"
```

It is site-wide. Frappe resolves every asset in the PDF pipeline against `frappe.utils.get_url()`, which on that site is `http://one_fm.15` — not a resolvable host — so wkhtmltopdf cannot fetch the print stylesheet either, and any PDF dies. Setting a reachable host fixes it outright:

```
site_config.json:  "host_name": "http://127.0.0.1:8000"

Employee / Standard                ->  PDF OK
Proof of Work Letter               ->  PDF OK
Proof of Work Attendance Report    ->  PDF OK
```

That is local configuration, not code, so it is not in this PR. Inlining the logo is still worth having: it removes one network dependency from a document that renders in a background job, and it is what puts the logo back on staging, where the fetch fails softly and leaves an empty box.

## RTL, second pass

`dir="auto"` alone was not enough — the values still carried their own `text-align`, so a Latin value hugged the left edge of an RTL table. Values now start at the cell's own edge, the right, with `unicode-bidi: plaintext` keeping each script in its own reading order: Arabic right-to-left, Latin left-to-right, both sitting where an RTL layout puts them.

## Housekeeping

An unrelated `patches.txt` line (`configure_agency_process_completion_values` commented out) was picked up from a working tree by an `add -A` in an earlier commit. It is restored, so this PR's net diff against `staging` is 7 files with no `patches.txt` change.

NoBoneZ (approved):

PR #6628, opened 2026-08-04
WI-001808: report Daily/Monthly work in days, and build the summary from the contract's items

yusuffGoodie:
Two fixes reported against [WI-001808](https://one-fm.com/app/work-item/WI-001872), both on the Proof of Work summary table (page 1 of the letter).

## 1. A Daily or Monthly item's work was reported in hours

The "Total number Days worked OR Total No of Hours worked" column rendered hours whatever the Rate Type. For Daily and Monthly items the hours were **days × the shift length**, so a month with 344 present days printed `4128.00 hrs` beside a contractual figure quoted as `690 DAYS` — two different units in the same row.

The column now splits on basis the same way the contractual figure and the staff breakdown either side of it already do, so all three of a row's numbers share one unit. The letter's worked cell renders the `OR` line like its neighbours, since a "Both" basis now puts two figures there.

Verified against every POW on the bench:

| Contract | Contractual | Worked (before → after) |
|---|---|---|
| Al Arabia Educational | `690 DAYS` | `4128.00 hrs` → **`344 Days`** |
| Al Babtain — Tea Boy | `60 DAYS` | `528.00 hrs` → **`44 Days`** |
| Al Babtain — Security Guard | `19344 HOURS` | `31860.00 hrs` (unchanged) |
| Tamdeen — Low Current Tech | `150 DAYS` | `1800.00 hrs` → **`150 Days`** |

## 2. Two rows where the contract has one line

Rows came from the union of Contract Items and items with attendance. They now come from the Contract Items alone.

**This change does not fix the underlying cause, and the reporter chose it knowing that.** The Alghanim Industries contract's only Service line is `UNF-TRS-000208-34-BLK-M` — a Tactical Trouser — and `SER-SEC-000136-NKW-M-30DY-12HR` (Security Guard), which 50 staff actually worked across 1,322 attendance records in January, is **not on the contract at all**. So that contract's summary is now a single row reading `90 DAYS` contracted against `0 Days` worked, and the security work no longer appears on page 1.

The contract is what needs correcting. It is 1 of 83 contracts carrying a Service line that is not a service item (the other, The Story Tower's `RCBT-NKW-M-30DY-12HR`, looks like a genuine service that simply lacks the `SER-` prefix).

Blast radius, checked by regenerating every POW on the bench: **only the two Alghanim documents change** (2 rows → 1). All others are byte-identical. The attendance sheet on the later pages is unaffected — it is built from attendance, not from the contract, and still lists every employee who worked.

## Verification

- `bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work` — **78 passed** (was 65; +13 covering both fixes)
- New `TestTheWorkedColumnUsesTheRightUnit` pins each basis, the nominal-hours path for Hourly items, half days, and that a row's two figures always agree in unit
- New `TestOnlyContractItemsBecomeRows` pins that an item with attendance but no contract line is not a row, and that a contract item with no attendance still is

## Note for review

Worth deciding whether a Contract Item typed `Service` whose item is not a service should be rejected at source — a validation on Contracts would stop this recurring silently. Say the word and I'll add it.


yusuffGoodie:
Pushed `635d081ca` — the attendance sheet now follows the contract too.

Restricting the summary rows left the sheet on the later pages still carrying a section for an item the summary no longer listed, so the two halves of one document disagreed about what was worked. `get_pow_attendance_report` now applies the same filter.

For the Alghanim Industries contract that means **no sections at all**. The print format already covers that case — it renders "No attendance recorded for this Proof of Work." rather than a bare page, so no template change was needed.

Blast radius, re-checked by running the report for every POW on the bench:

| Contract | Contract items | Sheet sections after |
|---|---|---|
| Al Babtain | `SER-FMG-000294`, `SER-SEC-000136` | both kept (2 and 93 employees) |
| Al Arabia | `SER-SEC-000136-NKW-M-26DY-12HR` | kept (25) |
| Kuwait Tennis Federation | `SER-CLN-000001` | kept (2) |
| Tamdeen | `SER-FMG-000771` | kept (7) |
| **Alghanim Industries** | `UNF-TRS-000208-34-BLK-M` | **none** |

Only the Alghanim documents lose a section; every other contract keeps all of its sections, because their attendance items are all on their contracts.

Tests: **82 passed** (was 78). New `TestTheAttendanceSheetFollowsTheContract` pins that an item with no contract line gets no section, that a contracted item keeps its section, that a mixed grid drops only the uncontracted items, and that the print format still has something to say when nothing survives.

Worth restating that this contract will now produce a Proof of Work with an empty summary *and* an empty attendance sheet, while 50 staff worked 1,322 recorded shifts in January. The document is doing what was asked; the contract is what makes it empty.

NoBoneZ (approved):

PR #6634, opened 2026-08-06
WI-001830: hand a first civil ID application to the PRO

yusuffGoodie:
[WI-001830](https://one-fm.com/app/work-item/WI-001830) — the Overseas PACI workflow: a PRO applies on the government portal, and the civil ID expiry lands back on the Employee.

> **Stacked on #6633 (WI-001824), which is stacked on #6632.** Base is that branch so this diff shows only this work; both touch `paci.json`. GitHub retargets as each merges. Review order: #6632 → #6633 → this.

## What lands

| AC | Where |
|---|---|
| A New Application PACI opens in `Pending PRO` and is assigned to the PRO | `create_PACI` → `hand_to_pro`, plus the `PACI-PRO` assignment rule |
| PRO enters the reference and submits → `Pending by PACI` | new workflow state + transition |
| HR/Admin completes it → `civil_id_expiry_date` updates on the Employee | already existed (`set_New_civil_id_Expiry_date_in_employee_doctype`); now covered by a test |
| PACI Reference Number field | new `PACI Details` section + `paci_reference_number`, placed as on BA |

Workflow additions, verbatim from the BA site (Workflow `PACI`, exported 2026-08-06 10:35) — states `Pending PRO` (edit: PRO), `Pending by PACI`, `Rejected`, and:

```
Draft            --Save-->    Pending PRO       (PRO)
Pending PRO      --Submit-->  Pending by PACI   (PRO)
Pending by PACI  --Approve--> Completed         (GR Operator)
Pending by PACI  --Reject-->  Rejected          (GR Operator)
```

## Why the state is set in code and not by the workflow

The AC says the state is `Pending PRO` **when the document is created**. The transition that gets there is allowed to the **PRO role**, and these records are opened by the system on behalf of a submitted Preparation — no PRO is in the session, and Frappe rejects the jump with `WorkflowPermissionError: transition not allowed from Draft to Pending PRO`. So `hand_to_pro` writes the field.

Writing a field leaves the assignment rules unaware — they only fire on the document's own save — so `apply(doctype=…, name=…)` is called explicitly, the same way `accommodation_leave_movement.py` already does. **Without it the record sits in Pending PRO assigned to nobody**, which is why the test asserts the ToDo and its assignee rather than just the state.

## Two silent failures the patch had to survive

1. **`create_workflow` logs failures to the Error Log instead of raising** — a patch can report success having changed nothing. This one verifies all three states and all four transitions afterwards and throws with what is missing.
2. **That guard fired on the first run**, and here is what it caught: `Workflow State.style` is mandatory on this site, and `create_workflow_state` passes the fixture's value straight through. A state row with no `style` never gets its master record created, and the workflow save then dies with `LinkValidationError: Could not find … State: Pending by PACI`. The three new rows carry BA's styles (`Warning` / `Danger` / `Primary`).

The rule reuses the site's existing `Action PACI - Apply in PACI` Process Task (`P-TASK-2026-00061`, assignee Ambrin Akbar Ali) instead of creating a second one — BA's copy is `P-TASK-2026-00078`, same task and same employee under a different name. Who covers the tier stays the process owner's call.

## Verification

`bench run-tests --module one_fm.grd.doctype.paci.test_paci_pro_workflow --skip-before-tests` — **6 passed**

- a New Application PACI opens in `Pending PRO` **and** produces an open ToDo for the Process Task's user
- Renewal and Transfer still open in Draft
- the workflow carries all four transitions with the right roles, `Pending PRO` is PRO-editable, and the `Pending by PACI` master exists
- the assignment rule is enabled, Process-Task-based, conditioned on `Pending PRO`, and its task has an assignee
- the reference number field is visible and editable (the PRO types into it)
- completing a PACI writes `civil_id_expiry_date` onto the Employee

Patch run end-to-end on the bench: **exit 0**, workflow and rule verified in the database afterwards. `test_preparation_new_actions` (WI-001824) re-run: **9 passed**, unaffected.

## One note

BA's export has **two `Completed` rows** (one `allow_edit: System Manager`, one `Government Relations Operator`) — almost certainly a double-click on the BA form. I kept the single existing `System Manager` row rather than copy the duplicate, since duplicate state rows make the state's `allow_edit` ambiguous and would list Completed twice. If Completed is genuinely meant to become GR-Operator-editable, say so and it is a one-line change.

## To deploy

`bench migrate` — runs `add_paci_pro_workflow_and_rule` (registered in `patches.txt`) and syncs the PACI doctype.


yusuffGoodie:
## Added the second `Completed` row — my dedupe was wrong

I flagged BA's duplicated `Completed` state as an export artefact and applied only one row. Having read Frappe's code, it is **not** an artefact — it is the mechanism:

```js
get_document_state_roles: function (doctype, state) {
    let workflow_states = frappe.get_children(..., "states", { state: state }) || [];
    return workflow_states.map((d) => d.allow_edit);   // every matching row
}
```

`allow_edit` holds **one role per row**, so granting two roles on one state means two rows, and Frappe collects the `allow_edit` of every row matching the state. By deduping I quietly took a **Government Relations Operator's edit rights on a Completed PACI** away.

Fixed. Both rows are now in the fixture, identical apart from the role, and applied on the bench:

```
Completed  doc_status=1  allow_edit=System Manager
Completed  doc_status=1  allow_edit=Government Relations Operator
```

### The check that should have caught it

The patch verified states and transitions — neither of which was missing, which is exactly why this got through. It now verifies the edit roles too. **Mutation-checked:** removing the row makes the patch refuse to finish:

```
WI-001830: the PACI workflow was not updated. Missing states: none.
Missing transitions: none.
Roles that cannot edit Completed: ['Government Relations Operator'].
```

Plus two tests: both roles come back for `Completed`, and the two rows agree on `doc_status` and the status field they set — so one state cannot come to mean two different things.

`bench run-tests --module one_fm.grd.doctype.paci.test_paci_pro_workflow --skip-before-tests` — **8 passed**.

### For the record, on the original question

Everything else already matched BA. All 11 transitions are identical, including `Pending PRO --Submit--> Pending by PACI`; every other state's `doc_status`, `allow_edit` and `update_field` matched; `paci_reference_number` is on the form; the `PACI-PRO` rule is applied with the same condition and points at the same assignee (**HR-EMP-00775, Ambrin Akbar Ali**), even though the Process Task is named per-site (`P-TASK-2026-00061` locally vs `P-TASK-2026-00078` on BA).

The only remaining difference is the **order** of the state rows — BA lists `Pending by PACI` second, this fixture appends the new states at the end. `workflow_state` is a Link field, so nothing reads that order except the Workflow document itself.

yusuffGoodie:
## The work item was rewritten after this was built — AC3 and AC4 are new

Nothing was wrong with the implementation: **WI-001830 was edited on 2026-08-06 at 19:31**, and this branch was committed at **09:14** the same morning. The version I built to had three ACs; it now has five, and two of them did not exist:

| AC | Status |
|---|---|
| 1 · New Application saved from Draft → `Pending PRO`, assigned to the PRO | already shipped |
| 2 · `Pending PRO` + reference + Submit → `Pending by PACI` | already shipped |
| **3 · Reject from `Pending by PACI` → pop-up for a Rejection Reason → `Rejected`** | **new — now built** |
| **4 · `Rejected` → Reapply button, clears the reference, resets to Draft, keeps a parent link** | **new — now built** |
| 5 · `Pending by PACI` + invoice + Approve → `Completed` → Employee civil ID expiry updates | already shipped |

The **Reject transition** was already here — it came with the BA workflow — which is exactly why the action appeared on the form but nothing was asked for. The dialog is what was missing.

## What lands now

**AC3.** Rejecting from `Pending by PACI` prompts for a **PACI Rejection Reason** and stores it on the document. The dialog reads its options **off the field**, so it can only offer values the field will accept on save — the same approach as the Work Permit and Visa Request reject dialogs.

**AC4.** A rejected application gets a **Reapply** button that raises a fresh one: the candidate, category and Preparation come across; the reference number and the rejection reason do not; the state resets to Draft; and **`rejected_paci`** carries the parent reference link. The rejected record is left as the audit history, and the server enforces the same gate as the button.

All three fields are **the ones BA supplied**, in the positions BA has them:

```
rejection_details_section   Section Break  "Rejection Details"
paci_rejection_reason       Select         Incorrect Address / Incorrect Data
rejected_paci               Link → PACI    read-only
```

One note: BA gives none of them a `depends_on`, so the Rejection Details section shows on every PACI, including a Draft. The Work Permit equivalent hides it until the record is rejected — say the word and I'll gate it the same way here.

## Verification

`bench run-tests --module one_fm.grd.doctype.paci.test_paci_pro_workflow --skip-before-tests` — **14 passed** (8 + 6)

- the dropdown offers exactly `Incorrect Address` / `Incorrect Data`, and is settable after submit
- the dialog is wired to `Pending by PACI` and reads its options off the field
- **the state the dialog fires from can actually reject** — a dialog on an action the workflow does not offer would never open
- only a `Rejected` PACI can be reapplied; Draft / Pending PRO / Pending by PACI / Completed are refused
- the link field points at PACI, is read-only and `no_copy`
- reapplying keeps the candidate and category, clears the reference and the reason, resets to Draft, and leaves the rejected record's reason intact

Confirmed the run leaves no PACI records behind.

yusuffGoodie:
## AC5 fixed — and it uncovered a data problem well outside this work item

```
ValidationError: Marital Status cannot be "Single".
It should be one of "Unmarried", "Married", "Widow", "Divorce", "Unknown"
```

Nothing to do with civil IDs. A **Property Setter** (`Employee-marital_status-options`) narrowed the Employee `marital_status` options to a different vocabulary **without migrating the rows**:

| stored value | employees | accepted? |
|---|---|---|
| *(blank)* | 2,370 | – |
| **Single** | **1,121** | ❌ |
| Married | 809 | ✅ |
| Unmarried | 97 | ✅ |
| **Divorced** | **2** | ❌ |
| **Widowed** | **1** | ❌ |

**1,124 employees hold a value the field refuses**, so *any* full `employee.save()` on one of them throws — regardless of what the save was for.

### What I fixed (in scope)

`set_New_civil_id_Expiry_date_in_employee_doctype` has two branches, and only the second did a full `employee.save()`. The first already wrote through `frappe.db.set_value` — which is exactly why this error only appeared for an employee with **no Civil ID row yet**.

The second branch now writes the same way: the Employee Document row straight in, the expiry date with `set_value`, and nothing else dragged through validation. A raw child insert gets no ordering for free, so `idx` is worked out rather than left to tie at 0.

Verified on **HR-EMP-04237**, whose Marital Status is `Single`:

```
work_permit_expiry_date : 2027-05-24
→ civil_id_expiry_date  : 2027-05-24   ✅
→ Employee Document row : idx 1, valid_till 2027-05-24   ✅
```

### What I did not fix, deliberately

**The data.** Mapping `Single → Unmarried`, `Divorced → Divorce`, `Widowed → Widow` across 1,124 employees is a vocabulary decision and a production data change — it belongs with whoever narrowed the options, not inside a PACI work item. It is breaking much more than this.

**The same pattern elsewhere.** These save the Employee the same way and will fail on the same 1,124 employees:

- `one_fm/grd/doctype/residency/residency.py:157`
- `one_fm/grd/doctype/work_permit/work_permit.py:367` and `:389`

Worth a single patch that either migrates the values or restores the old option list. Happy to write it as its own PR — say which direction you want.

`bench run-tests --module one_fm.grd.doctype.paci.test_paci_pro_workflow --skip-before-tests` — **16 passed**, including one that picks an employee whose Marital Status is out of options and asserts both writes land.

NoBoneZ (approved):

PR #6636, opened 2026-08-06
WI-001982: hold an Operations Role's Sale Item to the project's contract

yusuffGoodie:
[WI-001982](https://one-fm.com/app/work-item/WI-001982) — rostering and billing should only be able to name a Sale Item the client actually contracted for.

Independent of the other Operations-017 branches — base is `staging`.

## AC-2 and AC-3 are already done

The Proof of Work half of this work item landed with **WI-001808**: `_populate_pow_items` builds its rows from `_contracted_count_by_sale_item(doc.contract)` — the Contract Items alone, per contract — and the non-manpower table is a separate path keyed on Contract Item Category, so it is untouched by the restriction, exactly as AC-3 requires. Existing coverage: `TestOnlyContractItemsBecomeRows` in `test_proof_of_work.py`. Nothing to add there.

So this PR is AC-1: the Operations Role end.

## What lands

- **The dropdown** — `sale_item` now uses a server-side link query returning only items on the project's `Active` contracts. Resolved server-side rather than as an `in` list built on the client, so it cannot go stale against a contract that changed while the form was open. The old `is_stock_item: 0` filter moved into the query and still applies either way.
- **The rule** — `validate_sale_item_against_contract` enforces the same thing on save. The AC asks for a filter that cannot be bypassed, and a Link field can still be set through the API, a data import or a paste.

"Active" is the contract's workflow state — the same test POW generation already uses to decide which contracts are billable.

## Two limits, both driven by what the site actually holds

I checked the live data before choosing where the rule bites:

| | count |
|---|---|
| Operations Roles with both a project and a Sale Item | 1,015 |
| …whose Sale Item is **not** on their project's Active contract | **422** |
| …of those, Active | **149** |
| …of those, whose project has **no Active contract at all** | 259 |

**1. The rule applies only when the pairing is new or changed.** Holding every later edit of those 422 to the rule would block a status change or a shift change on a role nobody is trying to re-bill — that is rostering down. Introducing or changing the item is the act the AC blocks, and that is what is checked (`is_new()` or `has_value_changed`). Both directions are pinned by tests: an untouched legacy role still saves; changing its item to an off-contract one is refused.

**2. A project with no Active contract stays unrestricted.** `get_contracted_sale_items` returns `None`, not an empty set, for that case — the rule presupposes a contract to read the items from, and 259 of those roles belong to projects that have none (`ONE FM - Head Office` among them). An empty set would leave them unable to add a post at all. If you would rather those projects be blocked outright, it is a one-line change — but it will stop post creation on every internal project until each has an Active contract.

Worth a look separately: `EQU-ESP-000001` shows up as the Sale Item on a lot of the non-conforming Active roles across unrelated projects (Head Office, KNCC Accommodation, T4 Airport), which looks like a default that was never replaced rather than 149 individual mistakes.

## Verification

`bench run-tests --module one_fm.operations.doctype.operations_role.test_operations_role --skip-before-tests` — **8 passed**

- contracted items are read off Active contracts only
- a project whose contract is Inactive is unrestricted (`None`, not empty)
- the dropdown offers only the contracted item; unrestricted when there is no Active contract
- an off-contract item on a new role is refused, with the message naming the item and project
- a contracted item is accepted
- a legacy role holding an off-contract item still saves untouched
- changing that role's item to an off-contract one is refused

## To deploy

`bench restart` for the Python, and `bench build --app one_fm` (or a hard refresh) for the client script. No migration.

### DIFF
--- a/one_fm/operations/doctype/operations_role/operations_role.js
+++ b/one_fm/operations/doctype/operations_role/operations_role.js
@@ -3,10 +3,14 @@
 
 frappe.ui.form.on('Operations Role', {
 	refresh: function(frm) {
+		// WI-001982: only the Sale Items the project's Active contract covers. The
+		// is_stock_item filter moved into the query, which applies it either way - a
+		// project with no Active contract still gets every non-stock item.
 		frm.set_query("sale_item", function() {
 			return {
-				"filters": {
-					"is_stock_item": 0,
+				query: "one_fm.operations.doctype.operations_role.operations_role.sale_item_query",
+				filters: {
+					project: frm.doc.project,
 				}
 			};
 		});
--- a/one_fm/operations/doctype/operations_role/operations_role.py
+++ b/one_fm/operations/doctype/operations_role/operations_role.py
@@ -55,7 +55,21 @@ class OperationsRole(Document):
 			return
 
 		contracted = get_contracted_sale_items(self.project)
-		if contracted is None or self.sale_item in contracted:
+
+		if contracted is None:
+			# No Active contract means there is nothing to bill this role against, so the
+			# item cannot be checked - and an unchecked item is what the rule exists to
+			# stop. Named separately because the fix is the contract, not the item.
+			frappe.throw(
+				_(
+					"Project <b>{0}</b> has no Active contract, so there are no Sale Items to "
+					"choose from. Activate the project's contract before setting a Sale Item "
+					"on its roles."
+				).format(self.project),
+				title=_("No Active Contract"),
+			)
+
+		if self.sale_item in contracted:
 			return
 
 		frappe.throw(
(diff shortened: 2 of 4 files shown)


yusuffGoodie:
## Fixed: an uncontracted project no longer accepts any item

You pasted an off-contract code onto an existing role and it saved. The rule was working — you hit **the exemption I built into it**, which I flagged in the PR body and which turns out to be the wrong call.

Reproduced on two real roles, before the change:

| role | project | result |
|---|---|---|
| `Cashier-Carpark-Airport Terminal 4-Evening-1` | T4 Airport (**has** an Active contract) | **BLOCKED** ✅ |
| `CampBoss-Security-KNCC Mahboula Accommodation-Day-2` | KNCC Accommodation Service (**no** Active contract) | **SAVED** ❌ |

A project with no Active contract was left unrestricted, reasoning that the rule presupposes a contract to read items from and that blocking would leave internal projects unable to add a post. But the AC says the filter must not be bypassable, and an unchecked item is exactly what it exists to stop.

**Both cases are refused now**, with separate messages, because they are fixed differently:

- off-contract item → *"Sale Item **X** is not on any Active contract for project **Y**. Pick one of the items the project is contracted for."*
- no Active contract → *"Project **Y** has no Active contract, so there are no Sale Items to choose from. Activate the project's contract before setting a Sale Item on its roles."*

The dropdown now offers **nothing** for such a project rather than every non-stock item — which is what let the paste through in the first place. (Before a project is chosen at all, the list stays unfiltered: a new role has no project until the shift is picked.)

Same two roles after the change: both blocked, each with its own message.

### Blast radius, so it is not a surprise

**56 of the 587** Active roles that carry a project and an item are on projects with **no Active contract** — 17 projects:

| project | Active roles | contracts on record |
|---|---|---|
| General Services | 42 | **0** |
| ONE FM - Head Office | 14 | **0** |
| Eureka | 8 | 1 (not Active) |
| Weatherford Oil Tool ME | 8 | 1 (not Active) |
| KNCC Accommodation Service | 6 | 1 (not Active) |
| …12 more | 1–4 each | mixed |

**Ordinary edits to those roles are untouched** — the rule still only fires when the Sale Item or the Project is set or changed, so a status or shift change still saves. What now needs an Active contract first is **creating** a role there, or **re-pointing** an existing one.

The two large ones (General Services, Head Office) have no contract record at all, so they will need one — or a decision to exempt internal projects explicitly, which I would rather do by naming them than by leaving a hole for every project that happens to lack a contract. Say which you prefer.

`bench run-tests --module one_fm.operations.doctype.operations_role.test_operations_role --skip-before-tests` — **11 passed**, including the reported case: an off-contract item on a project with no Active contract is refused, a contracted item is refused there too (nothing is billable against an inactive contract), and the dropdown offers nothing.

NoBoneZ (approved):

PR #6637, opened 2026-08-06
WI-001983: name the Letter's columns after their units, and straighten the report header

yusuffGoodie:
[WI-001983](https://one-fm.com/app/work-item/WI-001983) — four layout fixes across the two Proof of Work print formats. Independent of the other Operations-017 branches; base is `staging`.

## 3.1 — Dynamic contract headers, no static "OR"

Both figure columns were headed `<days wording>` **OR** `<hours wording>` whatever the document reported, so a letter quoting only days still asked the reader to pick a line. `pow_letter_headers` now reads the units off the rendered rows and heads each column with only what is in it:

| Document | Column heading before | after |
|---|---|---|
| Tamdeen (days only) | Contractual Number of days per month **OR** Contractual number of hours per month | Contractual Number of days per month |
| Al Babtain (a Daily item and an Hourly one) | same static pair | both wordings, on their own lines, **no OR** |

Read off the rows rather than recomputed from the contract on purpose: a Contract Item whose Rate Type changed after the document was generated cannot then relabel figures that were produced under the old one.

**The in-cell "OR" stays.** It appears only where a row genuinely holds two figures (a "Both" basis), so it is data, not the static phrasing the AC asks to drop — and the AC's scenario is scoped to the headers.

## 3.2 — Metadata hierarchy

`Project, Contract, Client, Report Period` → **`Client, Project, Contract`**. Report Period leaves this line because 3.3 moves it into the header.

## 3.3 — Header alignment and logo

Logo **26px → 44px**; the header is now a two-cell table — logo and `ATTENDANCE REPORT` on the left, `Report Period` right-aligned on the same horizontal line. A table rather than flex, because wkhtmltopdf does not lay flex out reliably.

## 3.4 — Column overlaps

`Working Days` and `Days Off` shared one `.c-num { width: 32px }` and both wrapped inside it, which is what ran the two headings together. Each of the three totals columns now carries its own class and width (46 / 38 / 38px), wide enough for its own heading.

## Verification

Rendered both print formats as HTML against **three real POWs** on the bench (PDF rendering fails locally for an unrelated reason — wkhtmltopdf's `disable-local-file-access` and the scrubbed logo URL):

| Document | Letter heading | Metadata order | Report Period | Totals columns |
|---|---|---|---|---|
| Tamdeen Khiran Mall | days only ✅ | Client, Project, Contract ✅ | in header ✅ | 3 distinct ✅ |
| Al Babtain | both, no OR ✅ | ✅ | ✅ | ✅ |
| Alghanim Industries | days only ✅ | ✅ | ✅ | ✅ |

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **89 passed** (was 85; +4 new, and one existing assertion updated for the new class list).

New: a days-only document is headed in days, an hours-only one in hours, one reporting both carries both headings and no "OR", an empty document keeps both headings (a column headed with no unit reads as a missing heading). Plus template assertions for the separated widths, the Report Period's new home, and the Client→Project→Contract order.

## One thing left on the table

The Letter's third column — `اجمالي عدد ايام عمل / Total Number of Days` — also hard-codes days and can hold hours for an Hourly item. I left it: that heading is bilingual, so naming hours there means writing the Arabic for it on a document a client signs. Give me the wording and it is a two-line change.

## To deploy

`bench migrate` — both print formats are standard, and their `modified` stamps are bumped so the import is not skipped.


yusuffGoodie:
## Guarded: the Letter no longer dies when the heading method is not loaded yet

Surfaced while testing #6638 — every Proof of Work PDF failed with:

```
frappe.exceptions.PrintFormatError: Error in print format on line 7:
'pow_letter_headers' is undefined
```

**Diagnosis:** a branch/database mismatch, not a defect in either PR. This print format is a *database* record and was already synced onto the bench from this branch; the code checked out was #6638, which does not carry the `hooks.py` registration for the method. Confirmed:

| | |
|---|---|
| DB print format `Proof of Work Letter` | calls `pow_letter_headers(doc)` on line 7 |
| `hooks.py` on the WI-001981 branch | 0 registrations |
| `hooks.py` on this branch | 1 |

On `staging` with both merged there is no mismatch.

**But the same gap exists on a real deploy**, which is why it is worth a guard rather than only an explanation: the print format arrives with the database at `bench migrate`, the method arrives with the app code and is only live once the process restarts — and the **background workers that render these PDFs recycle on their own schedule**. In that window every letter fails, for every contract, rather than one heading being stale.

The heading now falls back to naming both units — exactly what it said before this work item — and picks up the dynamic version as soon as the method is there.

Verified both ways against the template's own line, so the test asserts on what ships rather than a copy of it:

```
without the method -> Contractual Number of days per month / Contractual number of hours per month
with the method    -> Contractual Number of Days
```

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **92 passed**, including that the method is still registered in hooks (the fallback is insurance, not the plan).

yusuffGoodie:
## Headings now follow the contract's Rate Types

Rebuilt to the rule you gave, on all three figure columns:

| Contract's Rate Types | Heading |
|---|---|
| all Daily / Monthly | days wording only |
| all Hourly | hours wording only |
| a mix (or an item with no Rate Type) | both, with **OR** between them |

The units come from **the same decision the rows were built with** — `_basis_for_rate_type`, which maps Daily/Monthly → days and Hourly → hours — so a heading cannot describe a column as something its figures are not.

### Why your Al Babtain letter showed OR

It is not a fallback or a bug — that contract genuinely mixes them:

```
SER-SEC-000136-NKW-M-30DY-12HR   Rate Type: Hourly    → row reports 29124.00 hrs
SER-FMG-000294-NKW-M-22DY-9HR    Rate Type: Monthly   → row reports 38 Days
```

Two rows, two units, one column. So OR is the correct heading there — and that is exactly the "else" branch of your rule.

Verified by rendering **every** Proof of Work on the bench:

| document | Rate Types | heading |
|---|---|---|
| Al Babtain (Jan & Apr 2026) | Hourly + Monthly | days **OR** hours, all three columns |
| Aayan Leasing | Monthly | days only |
| Tamdeen khiran Mall | Monthly | days only |
| Alghanim Industries | Monthly | days only |
| Al Arabia Educational | Monthly | days only |

### One thing to check before merge

The breakdown column is bilingual, so its Arabic has to follow the unit too. I used **`اجمالي عدد ساعات عمل`** ("total number of work hours") beside "Total Number of Hours", mirroring the existing `اجمالي عدد ايام عمل`. That wording is mine and goes on a document a client signs — worth a second pair of eyes.

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **94 passed**, covering all-Daily/Monthly, all-Hourly, mixed, an item with no Rate Type, a document with no rows, and that only the breakdown column carries Arabic. The migrate-without-restart fallback is kept and now covers all three columns.

The attendance-report header fix from the last round is unchanged: the document name stays on one line.

NoBoneZ (approved):

PR #6638, opened 2026-08-06
WI-001981: upload the Proof of Work export to Google Drive

yusuffGoodie:
[WI-001981](https://one-fm.com/app/work-item/WI-001981) — exported Proof of Work PDFs land in a shared Drive folder organised by month, instead of a ZIP someone has to pass around.

Independent of the other Operations-017 branches; base is `staging`.

## What lands

| AC | Behaviour |
|---|---|
| 1.1 Direct upload | The bulk export renders each contract's merged PDF and uploads it to the configured shared folder — no ZIP |
| 1.2 Dynamic subfolder | Resolved (or created) as `"Month Year"` — `July 2026` — under that folder |
| 1.3 Success notification | Realtime message with the count and a link straight to the folder the PDFs are in |

Authenticated as **ONE FM's own service account** via `google_credentials.get_drive_service()` — the identity the Document Register already uses. Nothing new to authorise, but the configured folder must be shared with that account.

## Where the setting lives, and why not core Google Settings

The work item says "Google Drive Link shall be added in Google Settings Page". I put it on **ONEFM General Setting → Google → Proof of Work**, beside the Service Account JSON that uploads into it.

That is a deliberate reading. Core `Google Settings` holds the OAuth client and the Maps key — nothing to do with the service account doing this upload. And the folder is useless unless it is shared with that exact account: Drive answers `404 File not found`, not a permission error, when it is not, so the failure reads as a missing folder rather than a missing grant. Keeping the two on one page is what makes that checkable. Say the word if you meant core `Google Settings` literally and I will move it.

It accepts a share link or a bare id, read through the Document Register's `_drive_id` — one definition of "what a Drive link is" rather than a second regex to get wrong.

## Two judgement calls

**One subfolder per document, not per batch.** A batch is normally a single month, but nothing stops two periods going out together, and a July record filed under August is worse than one extra API call. The folder is looked up by name every time rather than cached anywhere — it is a place in someone's Drive and can be renamed, moved or recreated between one month and the next.

**The ZIP stays as the unconfigured path.** The AC says Drive *instead of* the ZIP, and that is exactly what a configured site gets. But the alternative for a site with no folder set is an export button that stops working until someone pastes a link, and the ZIP is what every site has today. `enqueue_pow_zip` returns `destination: "drive" | "zip"` and the list view reports whichever it gets, so the user is never told the wrong thing. Happy to make it a hard failure if you would rather it not be possible to export locally once Drive is live.

A document that fails to render or upload is logged and skipped, the same contract the ZIP build already kept — one bad contract must not cost the others their upload.

## Verification

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **88 passed** (+6)

Drive is mocked; what is under test is which calls are made and with what:

- the subfolder is named after the period the document covers, not the month the export ran (`2025-12-31` → `December 2025`)
- an existing `July 2026` folder is reused — `files().create` is never called
- a missing one is created with the right name, MIME type and parent
- a pasted share link (`.../folders/1AbC…?usp=sharing`) resolves to its id
- the PDF is uploaded into the **period** folder under its `<client> - <contract> - <MMM-YYYY>.pdf` name, and the notification carries the count and that folder's link
- a site with no folder configured still enqueues `_build_pow_zip`

Not covered by tests, and not testable here: the real Drive round trip. It needs a service-account key and a shared folder, so it wants one manual run on staging once the folder is configured.

## To deploy

`bench migrate` (the new field on ONEFM General Setting), `bench restart`, then paste the shared folder link into **ONEFM General Setting → Google → Proof of Work Drive Folder** and share that folder with the service account shown on the same tab.


yusuffGoodie:
## Fixed: the Drive folder now lives on Google Settings

You're right — the story says *"Google Drive Link shall be added in Google Settings Page"* and I put it on **ONEFM General Setting**. My reasoning was that the upload authenticates with the service account JSON held there, so the folder sat beside its credentials — but that is not what the work item asked for, and one setting split across two pages by my judgement is worse than following the story.

**Moved.** `pow_drive_folder_link` is now a **Custom Field on Google Settings** (core doctype, so a Custom Field rather than a JSON edit), created by a patch, and removed from ONEFM General Setting.

The credentials stay put: the upload runs as ONE FM's own **service account**, not as the OAuth client Google Settings configures, so the folder still has to be shared with that service account. The field description says so on the page itself.

### The patch does not make you re-enter it

If you already set the folder while testing, the patch carries it across, and never overwrites a value already set on Google Settings. Both directions verified on the bench.

Getting that read right took three attempts, which is worth recording:

- `frappe.db.has_column("ONEFM General Setting", ...)` → **TableMissingError** — a Single doctype has no table of its own
- `frappe.db.get_single_value(...)` → **"Field does not exist"** — by the time the patch runs, the field is gone from that doctype's meta
- `frappe.db.get_value("Singles", ...)` → **"Unknown column 'modified' in 'ORDER BY'"** — that table has no `modified` column, so the default ordering has to be turned off

It reads `tabSingles` directly with `order_by=None`. The stored row outlives the field, which is exactly what makes the old value recoverable.

### The test that was missing

`test_the_folder_is_read_from_google_settings` asserts the reader is called with `("Google Settings", "pow_drive_folder_link")` **by name** — so reading the wrong settings page fails in the suite rather than in your testing. Plus one that the field exists on Google Settings, and one that it is gone from ONEFM General Setting (two places to set one folder is one too many).

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **91 passed**.

### To deploy

`bench migrate` (creates the Custom Field and carries the value over), then `bench restart`.

yusuffGoodie:
## The 403 is a folder permission — and the error now says so

Diagnosed against the configured folder by querying Drive directly:

```
folder:  "Proof of Work"  (1WnFDnoIIbs8FeBd1jiudnhhjA0RWE7ZG)
owner:   y.oyedele@one-fm.com          ← a personal My Drive, not a Shared Drive
capabilities: canAddChildren = false
              canEdit        = false
```

The service account **can see** the folder (or it would 404) but has **read-only** access, so creating the `Month Year` subfolder is refused. Nothing wrong with the code path — it is the sharing.

### To fix it

Share the folder with **`gmail-erpnext@inbound-theory-254808.iam.gserviceaccount.com`** as **Editor**.

**But prefer a Shared Drive.** That folder is in a personal My Drive, and a service account has **no storage quota of its own** there — so once it can write, uploads can still be refused with `storageQuotaExceeded`. On a Shared Drive the files belong to the drive, not the account, and the problem does not arise. The story asks for a "shared Google Drive folder", which is what that means in practice. The code already passes `supportsAllDrives=True`, so a Shared Drive needs no change.

### What changed in the code

The folder is now checked **once, before anything is rendered**, and the failure explains itself instead of raising a traceback per contract. Verified against your actual folder:

> **gmail-erpnext@inbound-theory-254808.iam.gserviceaccount.com** can see the Drive folder "Proof of Work" but cannot write to it, so the month's subfolder cannot be created. Share the folder with that address as **Editor** (Content manager on a Shared Drive).
>
> Note this folder lives in a personal My Drive. A service account has no storage of its own there, so uploads can still be refused for quota even once it can write — a **Shared Drive** is the arrangement that works.

and for a folder that is missing or not shared at all:

> The Proof of Work Drive folder (…) is not there, or is not shared with **gmail-erpnext@…**. Share it with that address, or correct the folder on Google Settings.

The service account address is not visible anywhere in the UI, which is most of why the original error was unactionable.

`bench run-tests --module one_fm.one_fm.doctype.proof_of_work.test_proof_of_work --skip-before-tests` — **94 passed**, including a read-only folder being refused with all three facts in the message, a My Drive folder being called out, and a writable Shared Drive folder passing.

NoBoneZ (approved):

PR #6639, opened 2026-08-06
WI-001979: make the attendance sheet's summary columns reconcile

yusuffGoodie:
[WI-001979](https://one-fm.com/app/work-item/WI-001979) — the summary block at the end of the Operations Monthly Attendance Sheet, so payroll can reconcile a row without manual math.

Base is `staging`. **WI-001980 will stack on this branch** — both change the same counting loop.

## The block

| Column | Counts |
|---|---|
| **Present Days** (was "Working Days") | `Present`, `Working`, `Work From Home` — days actually present |
| Days Off | `Day Off` + `Client Day Off` combined |
| **Absent Days** *(new)* | `Absent` |
| **Annual Leave** *(new)* | `On Leave` where the record's Leave Type is Annual Leave |
| **Sick Leave** *(new)* | …Sick Leave |
| **Leave Without Pay** *(new)* | …Leave Without Pay |
| **Other** *(new, see below)* | a status no column above names |
| **Missing/Empty Days** *(new)* | no attendance record at all for that day |

`summary_counter` is total rather than a chain of ifs with a gap at the end, so **the AC's reconciliation holds by construction**: every day increments exactly one counter, and the row always adds up to the days in the range.

AC-2 needed no code — `off_days` already summed both kinds of day off. It has a test now.

## The Other column, which the ACs do not mention

The AC asks that Present + Day Off + Absent + Leaves + Missing equal the days in the range. On this site's data that cannot hold, because some days are none of those. July 2026:

| status | days |
|---|---|
| On Hold | **288** |
| Client Interview | 5 |
| `On Leave` on a type outside the three named (Business Trip) | 2 |

Those days are recorded, so calling them Missing would be false, and leaving them uncounted breaks the very check the column block exists to provide. `Other` is where they go, and the identity holds. If you would rather have a column per leave type instead, note this site has eleven of them.

## Two smaller decisions

- **Leave type comes off `Attendance.leave_type`**, tracked in a map keyed by day alone. A leave covers the whole day and the report already mirrors it onto every one of the employee's shift rows, so its type cannot belong to one shift.
- **`working_days` and `off_days` keep their fieldnames.** The print format renders both by name; renaming would blank those cells rather than relabel them. Only the *label* changed, which is what the AC asks for.

## One thing left out, deliberately

The **print format** (`Operations Monthly Attendance Sheet`) still prints only Working Days and Days Off, under a hand-built fixed 44-column grid. Adding six columns there means re-doing that layout and its colspan arithmetic — not what this work item asks for, and easy to get subtly wrong on a page clients sign. Say the word and it is its own change.

## Verification

`bench run-tests --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests` — **57 passed** (48 existing + 9)

- each status maps to the column it should, and Present Days no longer absorbs Absent / On Leave / On Hold / Holiday
- both kinds of day off share one column
- each of the three leave types counts under its own column; a fourth type is Other, not Missing
- On Hold / Holiday / Client Interview are Other; a day with no record is Missing
- **every status in the Attendance doctype's own options list lands on a declared column** — so a new status cannot silently vanish from the row
- a constructed 7-day row reconciles exactly (1 present, 1 absent, 1 sick, 1 other, 2 off, 1 missing = 7)
- **and it reconciles on real data**: 200 rows of the busiest month in the database, each summing to the days in the range

One existing test (WI-001790's column allow-list) was updated to include the new fieldnames.

## To deploy

`bench restart`. No migration — report columns are code.

### DIFF
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
@@ -27,6 +27,64 @@ BY_SHIFT_HOURS = "Shift Hours"
 OVERTIME = "Over-Time"
 BASIC = "Basic"
 
+# WI-001979: the summary block at the end of a row. Every day in the range increments
+# exactly one of these, which is what makes them reconcilable - add them up and you get
+# the days in the range, with no cell counting.
+#
+# `working_days` and `off_days` keep their fieldnames (the print format renders them by
+# name); only the Present Days label changed.
+SUMMARY_COUNTERS = (
+	"working_days",
+	"off_days",
+	"absent_days",
+	"annual_leave_days",
+	"sick_leave_days",
+	"leave_without_pay_days",
+	"other_days",
+	"missing_days",
+)
+
+# Statuses that count as a day present. Work From Home and Working already showed as "P"
+# in the day cells, so they count here too - the column follows the cell.
+PRESENT_STATUSES = ("Present", "Working", "Work From Home")
+
+# Both kinds of day off share one column, per the AC.
+DAY_OFF_STATUSES = ("Day Off", "Client Day Off")
+
+# Leave Type -> its own column. A leave of any other type (Business Trip, Hajj Leave)
+# counts as Other: naming it would mean a column per Leave Type on this site, which is
+# eleven of them.
+LEAVE_TYPE_COUNTERS = {
+	"Annual Leave": "annual_leave_days",
+	"Sick Leave": "sick_leave_days",
+	"Leave Without Pay": "leave_without_pay_days",
+}
+
+
+def summary_counter(status, leave_type=None):
+	"""Which summary column one day belongs to (WI-001979).
+
+	Total, rather than a chain of ifs with a gap at the end: a day with a status this
+	does not name is Other, and a day with no status at all is Missing. Nothing falls
+	through uncounted, so the row's figures always reconcile to the range.
+	"""
+	if not status:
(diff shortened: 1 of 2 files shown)


yusuffGoodie:
## Follow-up: attendance with no Operations Shift was being dropped

Found while testing — reported against **HR-EMP-04101** over 24 Dec 2025 → 23 Jan 2026: **six** `Absent` records in the data, **five** on the report.

The sixth (14 Jan) has **no Operations Shift**, and both source queries *inner*-joined that table. The record never reached the report at all — so the day it recorded silently became a **"missing day"** instead of an absence.

It is not one record:

> **42,840 submitted attendance records in 2026 alone carry no Operations Shift**, and every one was invisible here.

That also made this PR's central promise false: a day that *was* recorded was being counted as missing, so the row reconciled to the range only because the dropped day landed in the wrong bucket.

Both joins are now `left_join`. Nothing else needed changing — the row key already read a missing shift as `""`, and the hours map only fills when a duration is present.

For that employee the row now reads **3 present / 6 absent / 1 other / 21 missing = 31**, matching the raw data exactly.

**Note:** these records group under a blank shift, so *on this branch alone* an employee can gain a second row. #6640, stacked on this, is where an employee becomes one row.

Tests added (60 in the suite here):
- the day a shift-less record covers is not reported as missing
- it lands in the counter its status belongs to
- the query is pinned to `left_join` — an inner join here loses rows silently rather than failing

NoBoneZ (approved):

NoBoneZ (approved):

PR #6640, opened 2026-08-06
WI-001980: an Include Project toggle that decides whether rows split by project

yusuffGoodie:
[WI-001980](https://one-fm.com/app/work-item/WI-001980) — an "Include Project" checkbox that controls whether an employee gets one row or one row per project.

> **Stacked on #6639 (WI-001979)** — both change the same counting loop, so the base is that branch. GitHub retargets to `staging` when #6639 merges.

## What it does

| Include Project | Rows | Project column |
|---|---|---|
| **off** (default) | one per employee (still split by shift and Day Off OT, as WI-001790 asks) | hidden |
| **on** | one per project the employee worked on, each carrying only that project's attendance | shown |

The summary counters from WI-001979 recount per row either way, so a project row's Present / Absent / Leave figures are that project's.

## Three things worth knowing

**The project now comes off the attendance, not the employee.** The Project column used to show `Employee.project` — a single current posting, which is not necessarily where the days in the row were worked. Both source queries (Attendance and Employee Schedule) now select their own `project`, and it joins the row key when the toggle is on.

**The row key keeps four slots either way**, with the project slot empty when the toggle is off, so everything that unpacks it reads one shape rather than branching on arity.

**The column is hidden, not dropped.** The client formatter colours day cells by column index (`colIndex < FIXED_COLUMNS`) and counts hidden columns — the report already carries a hidden `employee` column inside that count. Dropping a column would shift every day cell out from under the colour map. A test pins that both runs return the same number of columns.

**Whole-day events are unchanged and worth stating**: a leave or a day off is mirrored onto every one of an employee's rows, exactly as it already is across their shift rows. So with the split on, one day off appears on each project row. That is inherent to a row-per-scope report, not introduced here — but it means cross-row totals of days off are not a headcount of days.

## Verification

`bench run-tests --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests` — **64 passed** (57 + 7)

- two projects share one row key when the toggle is off and get one each when it is on
- the key has four slots either way
- the Project column is `hidden` when off, visible when on, and the column count is identical
- against the busiest project-month in the database: splitting **adds** rows and loses no employee, a split row carries its project, and **every split row still reconciles** to the days in the range (WI-001979's arithmetic survives the regrouping)
- splitting never loses a day worked, compared as per-employee totals

One note on that last test — it started life comparing row-to-row and failed on `HR-EMP-00007`: 30 vs 31. That was the test's fault, not the report's. The employee has two rows from the **Day Off OT** flag (30 present + 1), so a per-employee sum can never equal a single row. Rewritten to compare totals, with `assertGreaterEqual` because the two runs legitimately differ where one date carries attendance under two projects — consolidated they collapse onto one cell and count once; split, each project counts its own, which is the point of a per-project row.

## To deploy

`bench restart`, and `bench build --app one_fm` (or a hard refresh) for the new filter. No migration.

### DIFF
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.js
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.js
@@ -97,6 +97,14 @@ frappe.query_reports["Operations Monthly Attendance Sheet"] = {
 			label: __("Include Future Attendance"),
 			fieldtype: "Check",
 		},
+		{
+			// WI-001980: off, an employee gets one row and the Project column is hidden.
+			// On, they get a row per project they worked on. Not an in-page filter - it
+			// changes how the rows are grouped, so the report has to be re-run.
+			fieldname: "include_project",
+			label: __("Include Project"),
+			fieldtype: "Check",
+		},
 	],
 	formatter: function (value, row, column, data, default_formatter) {
 		value = default_formatter(value, row, column, data);
--- a/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
+++ b/one_fm/one_fm/report/operations_monthly_attendance_sheet/operations_monthly_attendance_sheet.py
@@ -61,6 +61,36 @@ LEAVE_TYPE_COUNTERS = {
 }
 
 
+# Which status stands when an employee has more than one record for the same day - two
+# shifts, or a basic shift beside day-off overtime. Presence wins: they worked that day,
+# whatever else was recorded against it. Anything not listed here ranks last, so a
+# recognised status always beats an unrecognised one.
+STATUS_PRECEDENCE = (
+	"Present",
+	"Working",
+	"Work From Home",
+	"Half Day",
+	"On Leave",
+	"Day Off",
+	"Client Day Off",
+	"Absent",
+)
+
+
+def more_significant(status, existing):
+	"""Should `status` replace `existing` as the day's cell?"""
+	if not existing:
+		return True
+
+	def rank(value):
+		try:
+			return STATUS_PRECEDENCE.index(value)
+		except ValueError:
+			return len(STATUS_PRECEDENCE)
+
+	return rank(status) < rank(existing)
+
+
 def summary_counter(status, leave_type=None):
 	"""Which summary column one day belongs to (WI-001979).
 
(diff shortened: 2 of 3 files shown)


yusuffGoodie:
## Fixed: an employee was still appearing on several rows with Include Project unchecked

You were right, and my reading of the AC was too narrow. The toggle only decided whether the **project** joined the row key — **shift, roster type and the Day Off OT flag still split it**, which is what put M D Abu Taleb on two lines and Abdoh Abdulaal on three in your screenshot.

The key is now the **project alone**: empty when the toggle is off (one row per employee), the project when it is on (one row per project they worked on).

Verified over your exact range, 24 Dec 2025 → 23 Jan 2026, across the whole report:

| | rows | employees | max rows for one employee | rows that don't reconcile |
|---|---|---|---|---|
| **Include Project off** | 1,560 | 1,560 | **1** | 0 |
| **Include Project on** | 1,814 | 1,560 | 5 | 0 |

### Two consequences worth knowing

**A day can now carry more than one record** — a basic shift beside day-off overtime, or two shifts. The cell is *decided* rather than overwritten by whichever the query returned last: **presence wins**, so a day worked shows as worked whatever else was recorded against it. That is also what keeps Present Days counting days actually worked, and it means a day is still one day in the summary.

**Roster Type and Day Off OT used to be carried by the row key**, so every row had one of each. They are now collected from the records behind the row and shown **only where those agree** — a row spanning Basic and Over-Time leaves the column blank rather than naming whichever was read last. Filtering by either makes them single-valued again, which is when they actually say something; there is a test pinning that. If you would rather they always show something, say the word.

### Also in this branch (from #6639, rebased in)

**HR-EMP-04101's absent count.** Six `Absent` records in the data, five on the report — the sixth had no Operations Shift and both queries inner-joined that table, so it was dropped entirely and became a "missing day". **42,840 submitted records in 2026 alone** have no shift. Both joins are now left joins; that employee now reads **3 present / 6 absent / 1 other / 21 missing = 31**, matching the raw data.

`bench run-tests --module one_fm.one_fm.report.operations_monthly_attendance_sheet.test_operations_monthly_attendance_sheet --skip-before-tests` — **77 passed**, including new tests that an employee is exactly one row when the project is not included, one row per project when it is, the single row still reconciles, and the day-cell precedence rules.

NoBoneZ (approved):

PR #6642, opened 2026-08-06
WI-001976: reapply a Visa Request PAM rejected over the designation or the gender

yusuffGoodie:
[WI-001976](https://one-fm.com/app/work-item/WI-001976) — when PAM rejects for something the application can be corrected for, raise a fresh attempt without re-keying the applicant.

Independent of the other Operations-017 branches; base is `staging`.

## What lands

A **Reapply Visa** button on a `Rejected By PAM` request whose PAM reason is one of two. It creates a new request carrying the applicant, named with the next number in the original's series (`VR-08-2026-00002-1`), and leaves the rejected one untouched as the history.

## No new rejection-reason field was needed

The AC names the two reasons as "Designation" and "Gender (The worker's gender does not match the profession)". Those map onto the predefined PAM reasons the Reject dialog **already** offers (WI-001693) and writes into `pam_rejection_remark`:

| Work item | Existing reason |
|---|---|
| Gender | `The worker's gender does not match the profession` — the AC quotes it verbatim |
| Designation | `The occupation requires amendment to specify the worker's specialization` |

The button stays away from `Worker is in Black List` and `An active file exists for this worker`, where a fresh request would be refused for the same cause. The server enforces the identical gate — a button is a convenience, not a permission.

> The Designation mapping is my reading: it is the only occupation/designation reason in the list. Worth a nod before merge.

## The amendment series, without Frappe's amendment machinery

`amended_from` may only point at a **cancelled** document. A `Rejected By PAM` request is still a draft — the workflow gives that state `docstatus 0` — so it can be neither submitted nor cancelled to qualify. Frappe refuses it outright, which I confirmed on a real record:

```
ValidationError: VR-08-2026-00002 cannot be amended because it is not cancelled.
```

So the number comes from Frappe's own `append_number_if_name_exists`, and a new **`reapplied_from`** link carries the audit trail — the same shape the BA site chose for the Work Permit equivalent (`rejected_work_permit`). If you would rather have the real amendment chain, that means making `Rejected By PAM` a submitted state in the workflow, which is a bigger change than this work item and affects every existing rejected request.

**The new request is a copy, not a blank with two fields set.** `passport_copy` is mandatory and has nothing to fetch from, so a bare new request cannot be saved at all. What does *not* carry over is the outcome of the failed attempt — every rejection remark, reference number, decision date, the visa document and the payment receipt — and the state resets to Draft.

## Verification

`bench run-tests --module one_fm.visa_management.doctype.visa_request.test_visa_request --skip-before-tests` — **7 passed**, against a real Visa Request on the bench

- both reasons allow a reapplication; Black List, active file and a blank reason do not
- a request in any other state cannot be reapplied
- the new name is the original's plus a number, and a second attempt increments it — reapplying a reapplication gives `-2`, not `-1-1`
- the new request carries Job Offer, Job Applicant, passport number and passport copy, and links back through `reapplied_from`
- it starts in Draft with **every** outcome field cleared (asserted over the list itself, so a field added to it later is covered)
- the rejected request keeps its state and docstatus
- the server refuses a request that does not qualify, independently of the button

Confirmed the run leaves nothing behind: no `Visa Request` with `reapplied_from` set remains afterwards. One thing to know for future tests here — `FrappeTestCase` rolls back per **class**, not per test, so a record created by an earlier test in the same class is visible to the next one. The naming test is written to be order-independent because of it.

## To deploy

`bench migrate` (the new `reapplied_from` field), `bench restart`, and `bench build --app one_fm` for the button.


yusuffGoodie:
## Fixed: this PR was carrying WI-001873's commit

You were right to query it. I branched `WI-001976-reapply-visa` off **`WI-001873-resumption-details`** instead of off `staging`, so the Leave Application resumption work rode along in a PR whose base is `staging`. Nothing wrong with either change — they were simply in the wrong place.

Rebased onto `staging` with `--onto`, dropping the WI-001873 commit. This PR is now one commit and four files, all Visa Request:

```
one_fm/visa_management/doctype/visa_request/visa_request.py
one_fm/visa_management/doctype/visa_request/visa_request.js
one_fm/visa_management/doctype/visa_request/visa_request.json
one_fm/visa_management/doctype/visa_request/test_visa_request.py
```

The resumption work is unaffected and still on **#6641**, where it belongs.

`bench run-tests --module one_fm.visa_management.doctype.visa_request.test_visa_request --skip-before-tests` — **7 passed** on the rebased branch.

### I audited the rest of the sprint's branches for the same mistake

Only this one was wrong. Every other multi-commit branch is a **deliberate stack whose PR base is the parent branch**, not `staging`:

| branch | carries | PR base |
|---|---|---|
| WI-001824 | + WI-001973 | #6632 ✅ |
| WI-001980 | + WI-001979 | #6639 ✅ |
| WI-001829 | + WI-001974 | #6644 ✅ |
| WI-001827 | + WI-001829, WI-001974 | #6645 ✅ |
| WI-001828 | + WI-001827, WI-001829, WI-001974 | #6646 ✅ |

Those extra commits are expected — GitHub shows only each PR's own diff and retargets to `staging` as the stack merges.

NoBoneZ (approved):

PR #6643, opened 2026-08-07
WI-001977: read the Visa Copy and Payment Receipt into the Visa Request

yusuffGoodie:
[WI-001977](https://one-fm.com/app/work-item/WI-001977) — OCR the two documents the GRD Operator attaches once a visa is issued, so the four fields are not keyed in by hand.

Unblocked by the model ids added to `site_config.json` (`e-visa_model_id`, `receipt_model_id`) and the sample responses supplied with the work item. Base is `staging`.

## The mapping, as the work item specifies it

| Attachment | Mindee field | Visa Request field |
|---|---|---|
| Visa Copy (`visa_document`) | `reference_number` | Visa Reference Number |
| Visa Copy | `date_of_issue` | Visa Issue Date |
| Visa Copy | `date_of_expiry` | Visa Expiry Date |
| Payment Receipt (`payment_receipt`) | `date` | Payment Date |

Written from the supplied samples and pinned by tests against them, so it reflects what the models actually return rather than what they might. Run over the sample e-visa that is `385515182` / `2026-01-19` / `2026-04-18`, and over the sample receipt, `2026-01-19`.

Two oddities in the samples the mapping has to survive, both covered:
- **the reference number arrives as a number** and the target is a Data field — cast through `int()` first, or a float would leave a `.0` on the end (the same trap the Civil ID parser documents)
- **fields the model could not read come back with a `value` of `null`** (`place_of_birth` in the sample) — those are left out rather than written as blanks, so a partly-read visa is still worth presenting

## How it runs

- **Enqueued, not inline.** Mindee is an external call taking seconds; a save should not wait on it. The form picks the values up over realtime and shows which fields were filled — without that the operator would be looking at a stale form and would type them in anyway.
- **Only in `Pending Visa Issuance`**, which is the state the AC names. A passport copy attached to a Draft has nothing to do with a visa.
- **Only an attachment that actually changed**, so a correction the operator makes to an extracted date is not overwritten by the next save.
- **One unreadable document does not cost the other its extraction** — it is logged and the field can still be filled by hand.
- Nothing advances the workflow. The AC asks for the values to be reviewable before the next action, and they are ordinary editable fields.

A missing model id **throws** naming the config key rather than silently doing nothing — a deployment gap should read as one.

## Verification

`bench run-tests --module one_fm.tests.test_visa_document_ocr --skip-before-tests` — **15 passed**

- both mappings produce exactly the expected values from the supplied samples
- a numeric reference does not arrive with a decimal tail
- nothing is taken from the visa that was not asked for (it also carries a passport number and a date of birth)
- a field the model could not read is omitted; a response missing it entirely does not raise
- **every target fieldname exists on Visa Request** — a typo would write nowhere and read as OCR failing
- the trigger fires for an attachment added in the right state, handles both at once in one job, and stays silent for: no attachment changed, another workflow state, a cleared attachment
- a missing file on disk and an empty URL are reported rather than read
- each document names a real, importable extractor

**Not called against Mindee.** The work item supplied the responses, not the source documents, so there was nothing to send. Worth one attachment on staging after deploy to confirm the round trip; the parsing either side of the call is what the tests cover.

## To deploy

`bench restart`, plus `bench build --app one_fm` for the realtime listener. The two model ids are already in `site_config.json` on this bench — they need adding on staging and production.

### DIFF
--- a/one_fm/ocr_utils.py
+++ b/one_fm/ocr_utils.py
@@ -11,6 +11,7 @@ Uses the same Mindee implementation as the magic link module for consistency and
 
 import frappe
 import re
+from frappe import _
 from datetime import datetime
 from mindee import ClientV2, InferenceParameters, PathInput
 import os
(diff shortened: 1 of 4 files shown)


yusuffGoodie:
## Fixed — and the error *was* logged, just not where you looked

The job ran. It failed with this, against `VR-08-2026-00007`:

```
Visa Request OCR failed — Visa Copy
VR-08-2026-00007
  File ".../visa_request.py", line 168, in run_document_ocr
    file_path = _attachment_path(doc.get(fieldname))
  File ".../visa_request.py", line 195, in _attachment_path
    raise ValueError("No file attached")
```

Worth knowing for next time: `frappe.log_error(title=...)` writes into the Error Log's **`method`** column, not a `title` column — so the list view's default columns can make an entry look absent. The record was `s50vpa1uj4`.

### The cause

`on_update` runs **inside the save transaction**, and `frappe.enqueue` pushes the job immediately. A worker can pick it up and re-read the document **before that transaction commits** — seeing the row as it was, with no attachment on it. That is precisely the "No file attached" on a request that plainly had one.

The enqueue is now **`enqueue_after_commit=True`**, which is exactly what that flag is for.

An attachment that really has gone by the time the job runs — removed between the save and the worker, which your activity log shows happening — is now skipped quietly instead of raising through the traceback path. Nothing to read, nothing worth an Error Log entry.

### Verified end to end on your record

With the Mindee call stubbed (so the test is about the plumbing, not the model), attaching both documents to `VR-08-2026-00007` and running the job as the worker does:

```
visa_reference_number : 385515182
visa_issue_date       : 2026-01-19
visa_expiry_date      : 2026-04-18
payment_date          : 2026-01-19
```

The record was put back exactly as it was afterwards — attachments empty, your own 09-08-2026 / 11-08-2026 dates intact.

**Mutation-checked:** dropping `enqueue_after_commit` fails the new test. 17 tests pass.

### Separate, and not from this PR

Your bench also logged `Error Attaching File → FileNotFoundError: './one_fm.15/private/files/EVisa-283059338.pdf'`. That is Frappe failing to create a File record whose content is not on disk — the file had already been removed (there is no `EVisa-*` left in `private/files`). Re-attaching a fresh copy of the PDF should be clean.

NoBoneZ (approved):

yusuffGoodie:
Both review points applied in `e15fe5aa1`.

## 1. Visa Number, not Reference

```diff
-"reference_number": "visa_reference_number",
+"visa_number": "visa_reference_number",
```

Worth noting these are two different numbers on the same document — the Visa Number at the top (`283059338`, رقم التأشيرة) and the Reference below it (`385515182`, رقم المرجع). The old mapping was taking the second. A test now asserts the first is taken **and** that the second appears nowhere in the result, so a silent swap back would fail.

## 2. Payment Date is a Datetime

The receipt returns the two parts separately:

```json
"date": "2026-01-19",
"time": "19:36:05"
```

`payment_date` changed from `Date` to `Datetime` and now records both — two payments on the same day are distinguishable by their receipts. The time rides through the field mapping under a carrier key and is folded into `payment_date` afterwards, so nothing that is not a real Visa Request field ever reaches the document (writing one would raise on save). A receipt with no time still yields a usable date; it lands at midnight, which is what a Datetime does with a bare date anyway.

**Existing data:** the two Visa Requests that already carried a `payment_date` converted cleanly to `2026-01-19 00:00:00` on the column change — no loss.

21 tests, mutation-checked both ways: restoring `reference_number` fails three by name, dropping the time merge fails one.

PR #6645, opened 2026-08-07
WI-001829: capture why a Work Permit was rejected, in a field a report can read

yusuffGoodie:
[WI-001829](https://one-fm.com/app/work-item/WI-001829) — structured rejection reasons on Work Permit, for both PAM and the previous company.

> **Stacked on #6644 (WI-001974)** — the automatic rejection applies the `Reject` transition that PR adds, so it needs that base.

## The two dropdowns

| Field | Options (as BA now has them) |
|---|---|
| PAM Rejection Reason | `PAM Contract`, `Gender` |
| Previous Company Rejection Reason | `Auto rejected after 3 days`, `Previous Sponsor Rejected` |

Both rejections land in the **same `Rejected` state** (the BA's decision — no new states), so `reason_of_rejection` is now hidden and does the one job of saying which kind it was. Each dropdown shows only for its own kind, and the section only on a rejected permit — the AC's "revealed dynamically only when the respective rejection event occurs".

The old free-text `reason_of_rejection` / `details_of_rejection` are **hidden, not deleted**: they hold the history on permits already rejected.

## The dialog

Rejecting from `Pending By PAM` or `Pending By Previous Company` now asks for the reason on the way out and stamps both the reason and which kind of rejection it was. The dialog reads its options **off the field itself** rather than a list kept in the client — the same approach Visa Request took in WI-001693, and for the same reason: a hardcoded list only drifts from what the field will accept on save.

## Three working days, and why the weekend is hardcoded

Silence from the previous employer past three working days is a refusal, swept daily on the same working-days cron as the other GRD reminders. It **applies the workflow** rather than writing the state — `Rejected` is a submitted state, so a `db_set` would leave a permit reading Rejected while still at docstatus 0.

The weekend (Friday/Saturday) is counted as a floor in code, with public holidays taken off the company's Holiday List on top. That is deliberate, and this is why:

> the company's `default_holiday_list` on this database is **"Default Company Holiday List 2024"** — it runs to **2024-12-31** and has **no weekly off at all**.

A count that trusted it would treat every Friday since 2025 as a working day and expire requests early. There is a current list on the site (`Default Company Holiday List`, 2025→2040, weekly off Friday) that simply is not set as the company default — **worth fixing separately**, and this code will pick its public holidays up automatically once it is.

The clock starts at `inform_previous_company_on`, which the form already stamps when the previous company status is first set; permits predating that field fall back to when they entered the state.

## Verification

`bench run-tests --module one_fm.tests.test_work_permit_rejection_reasons --skip-before-tests` — **17 passed**

- both dropdowns offer exactly the listed reasons; each shows only for its own kind of rejection; neither shows on a permit that was not rejected
- a rejection cannot be left without a reason, and both are settable after submit (Rejected is submitted)
- the free-text pair they replace is hidden
- the dialog reads its options off the field
- **Wednesday → Saturday is 2 working days, not 3** (Thu counts, Fri and Sat do not)
- **the count does not depend on the stale holiday list** — a Thursday→Friday span is 0 working days even though the list returns no holidays for 2026
- the auto-rejection writes a reason the Select accepts and marks which kind it was
- the state it sweeps can actually be rejected (i.e. #6644's transition exists), and the job is registered on the cron

## To deploy

`bench migrate` (doctype), `bench restart` (cron), `bench build --app one_fm` (dialog).

### DIFF
--- a/one_fm/grd/doctype/work_permit/work_permit.js
+++ b/one_fm/grd/doctype/work_permit/work_permit.js
@@ -325,3 +325,61 @@ var set_required_documents = function(frm) {
   }
     frm.refresh_field('documents_required');
 };
+
+// WI-001829: a rejection has to say why. Both rejecting states land in the same
+// "Rejected" state, so the reason is asked for on the way out and the state it came
+// from is recorded in reason_of_rejection - which is what decides, afterwards, which of
+// the two dropdowns the form shows.
+const REJECTION_REASON_BY_STATE = {
+	'Pending By PAM': {
+		field: 'pam_rejection_reason',
+		rejected_by: 'Rejected by PAM',
+		title: __('PAM Rejection Reason')
+	},
+	'Pending By Previous Company': {
+		field: 'previous_company_rejection_reason',
+		rejected_by: 'Rejected by previous company',
+		title: __('Previous Company Rejection Reason')
+	}
+};
+
+frappe.ui.form.on('Work Permit', {
+	before_workflow_action: function(frm) {
+		if (frm.selected_workflow_action !== 'Reject') return;
+
+		const spec = REJECTION_REASON_BY_STATE[frm.doc.workflow_state];
+		if (!spec) return;
+
+		return ask_for_rejection_reason(frm, spec);
+	}
+});
+
+function ask_for_rejection_reason(frm, spec) {
+	// The options come from the field itself rather than a list kept here, which would
+	// only drift from what the field will accept on save.
+	const df = frappe.meta.get_docfield('Work Permit', spec.field, frm.doc.name);
+	const options = (df && df.options ? df.options : '').split('\n').filter((o) => o);
+
+	return new Promise((resolve, reject) => {
+		frappe.prompt(
+			[{
+				label: spec.title,
+				fieldname: 'reason',
+				fieldtype: 'Select',
+				options: options.join('\n'),
+				reqd: 1
+			}],
+			(values) => {
+				frm.set_value(spec.field, values.reason);
+				frm.set_value('reason_of_rejection', spec.rejected_by);
+				frm.save().then(resolve).catch(reject);
+			},
+			spec.title,
+			__('Reject')
+		);
+
+		// Frappe freezes the form for the workflow action; the prompt is unusable until
+		// it is released.
+		frappe.dom.unfreeze();
+	});
+}
--- a/one_fm/grd/doctype/work_permit/work_permit.js
+++ b/one_fm/grd/doctype/work_permit/work_permit.js
@@ -372,7 +372,14 @@ function ask_for_rejection_reason(frm, spec) {
 			(values) => {
 				frm.set_value(spec.field, values.reason);
 				frm.set_value('reason_of_rejection', spec.rejected_by);
-				frm.save().then(resolve).catch(reject);
+				// The reason has to be in the database before the action runs:
+				// apply_workflow() reloads the document server-side and throws away
+				// whatever the form was holding. Saving a document with nothing to save
+				// only shows "No changes in document" and never calls back, which would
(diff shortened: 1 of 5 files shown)


yusuffGoodie:
## Fixed a real bug this PR introduced — found while testing #6644

Rejecting a Local Transfer from `Pending By Previous Company` threw:

> **First, You Need to set the reason of Rejection Mentioned in PAM Website**
> Mandatory fields required in Work Permit form: **Reason Of Rejection**, **Details of Rejection**

…with no fields on screen to fill. To answer the question directly: **no, they are not hidden fields that get filled automatically — they were unfillable.**

### Why

`check_required_document_for_workflow` demands that pair whenever a Local Transfer reaches `Rejected`. It is pre-existing code, and #6644 simply opened a new door into it — but on **this** branch both fields are `hidden: 1`, because they are exactly the free-text pair the two rejection Selects replace. So the rejection could neither pass the check nor be filled in to satisfy it.

It is no better on the base branch: there they are visible, but only via `depends_on: workflow_state == "Rejected"` — the very state the check prevents reaching. Unfillable either way, which is the chicken-and-egg you hit.

### The fix

That check no longer asks for them. It still updates the Transfer Paper's child row, which is the rest of what that branch does. The reason is **still required**: the two Selects carry `mandatory_depends_on` tied to which kind of rejection it was, and the Reject dialog asks for it on the way out.

Verified on **WP-2026-00367**:

| | result |
|---|---|
| Reject with the dialog's values | → `Rejected`, reason stored as `Previous Sponsor Rejected` |
| The 3-working-day sweep | → `Rejected`, reason `Auto rejected after 3 days`, **0 error logs** |

That second row matters: the auto-rejection was hitting the same check and failing silently into the Error Log.

**Mutation-checked** — putting the pair back fails both new tests. `bench run-tests --module one_fm.tests.test_work_permit_rejection_reasons --skip-before-tests` — **19 passed**.

#6646 and #6647 have been rebased onto this so the whole stack carries the fix.

### One thing left open for you

A reject through the **API** (not the form) can still reach `Rejected` with no reason at all, because `mandatory_depends_on` only fires once `reason_of_rejection` names a kind. I did not force it globally on purpose: `Pending By Supervisor --Reject--> Rejected` is a pre-existing path with **no** dialog and no reason field of its own, and demanding one there would break it. If a supervisor rejection should also capture a reason, that is a new AC — say so and I will add it to the dialog and enforce it server-side for every path.

yusuffGoodie:
Follow-up from testing AC2 (reject at **Pending By PAM**, reason *PAM Contract*): the rejection could not leave the state at all. Two separate causes, both fixed in `2884de32c`.

**1. It demanded a payment invoice.** `validate_workflow_state_fields` requires `attach_invoice` and `new_work_permit_expiry_date` on *every* save made while the document sits in `Pending By PAM` — it cannot tell an Accept from a Reject. Nothing is paid on a rejection, so the reject dialog's save was blocked by a document that does not exist. The demand now applies only when there is no PAM rejection reason, which is exactly what the two Accept transitions look like.

**2. The retry then hung silently.** The dialog called `frm.save()` unconditionally. A save with nothing to save only shows a "No changes in document" alert and never calls back — so on the second attempt, where the reason was already stored from the first, the promise never settled and the action never ran. No transition, no error. It now saves only when the form is dirty.

The reason does have to reach the database before the action: `frappe.model.workflow.apply_workflow` calls `doc.load_from_db()` and discards whatever the form was holding, so the values cannot ride along with the action itself.

Verified on WP-2026-00372 with no invoice attached: the reason saves, `Reject` applies, and a plain save (no reason, i.e. heading for approval) is still blocked on the invoice. One new test, mutation-checked; 20 pass. #6646 and #6647 rebased on top.

NoBoneZ (approved):

PR #6646, opened 2026-08-07
WI-001827: make PAM's rejection a single step

yusuffGoodie:
[WI-001827](https://one-fm.com/app/work-item/WI-001827) — the PAM approval branch for an Overseas Work Permit.

> **Stacked on #6645 (WI-001829)** → #6644 (WI-001974) → `staging`.

## What lands

`Pending By PAM --Reject--> Rejected`, in one step. It used to park in **"Reason of Rejection"** and need a second click to become Rejected — a state a permit could be forgotten in, which told the reader nothing the reason field does not.

Per the BA, **"Rejected by PAM" is the existing `Rejected` state**, with the reason recorded in the dropdown #6645 adds — which is what the retired state existed to capture, in free text.

**BA's export carries this transition twice.** Applied once: two identical Reject buttons on one state is a UI bug waiting to happen.

## Nothing is stranded

Retiring the intermediate state leaves it with no transitions, so the patch moves what is in it onto `Rejected` **before** applying the workflow — there is never a moment where a permit holds a state with no way out. Two permits are in it on production (`WP-2023-00606`, `WP-2024-00420`), both rejections halfway through the old two-step flow.

The state stays *defined* so anything that somehow still holds it can be read. The patch refuses to finish if a permit is left there, if the state still routes anywhere, or if the new transition did not apply — `create_workflow` logs failures instead of raising, so it cannot be taken on trust.

## One AC I did not implement — needs your call

> **AC-1:** *Given an Overseas Work Permit in Pending PAM Approval, when PAM approves, then the application goes to Ready for Payment.*

BA's workflow does not do this. `Pending By PAM --Accept--> Pending  For Payment` is conditioned on **`doc.work_permit_type == "Local Transfer"`**, so an **Overseas** permit approved by PAM goes to `Completed` instead, exactly as it does today.

I left it alone rather than guess, because changing it is not a one-liner:

- adding Overseas to the payment branch means also **excluding** it from the unconditional `Accept --> Completed`, or the state offers two Accept buttons and which one wins is down to transition order
- that unconditional Accept is shared by every other type (both renewals, New Kuwaiti, Cancellation), so editing it reaches well beyond this work item

**Does an Overseas permit go through payment after PAM approval, like a Local Transfer?** If yes, I will condition both branches explicitly and add the pair of tests; it is a small change once the answer is known.

## Verification

`bench run-tests --module one_fm.tests.test_work_permit_workflow --skip-before-tests` — **11 passed** (6 from #6644 + 5), against both the fixture and the applied workflow

- PAM rejects straight to Rejected; the two-step route is gone from both
- the Reject action is offered exactly once
- no permit is left in the retired state
- a rejected permit can still say why (the #6645 dropdown is wired to the same discriminator)

Patch run on the bench: it moved the stranded permits and the verification passed.

## To deploy

`bench migrate`.


yusuffGoodie:
Adds the part of WI-001827 that was missing: **the Assignment Rules**. The work item links to three for Work Permit and none of them existed on this instance — only the unrelated `Work Permit Completion - GR Operator`.

| rule | assigned while | Process Task |
|---|---|---|
| `Work Permit-PRO` | Apply Online by PRO | Assigning PRO in Work Permit |
| `Work Permit - GRD Supervisor` | Pending By Supervisor, Pending  For Payment | Assigning GRD Supervisor in W |
| `Work Permit - GR Operator` | Apply Online by PRO, Pending By PAM, Pending By Previous Company, Pending By Operator | Set new Work Permit Expiry Date and mark as completed. |

Rule type is **Based on Process Task**, so the assignee comes from the task's employee — an unlinked rule assigns nobody. The patch creates each task (reusing one that already matches rather than inserting a second copy) and verifies afterwards, since `create_assignment_rule` logs failures instead of raising.

**Two departures from the export, both because it does not work as written:**

1. `Work Permit - GRD Supervisor` tests `doc.workflow_state`. `assign_condition` is evaluated as `frappe.safe_eval(condition, None, doc_dict)` — the document's own fields are the locals, so `doc` is not a name that exists. Verified against a real Work Permit: `NameError: name 'doc' is not defined`, which `AssignmentRule.safe_eval` swallows into a "Auto assignment failed" msgprint. The rule would never have fired and every save would have shown that message.

2. The export spells the payment state `"Pending For Payment"`; the workflow state is `"Pending  For Payment"` — **two spaces**. As written the supervisor is never assigned while a transfer waits for payment. Both spellings are in the condition so it works either way.

**Workflow:** the fixture on this branch already matched BA's export exactly, states and transitions, with one exception — BA carries `Pending By PAM --Reject--> Rejected` twice. That row is now duplicated here too. The two are identical, so `apply_workflow`'s last-match-wins picks the same transition and `add_dropdown_item` drops the repeat because it keys on the label.

Applied on this instance and re-run to confirm idempotency (no duplicate tasks). 8 new tests + 11 workflow tests, mutation-checked — restoring BA's verbatim `doc.workflow_state` fails the suite with the NameError.

yusuffGoodie:
Correction to my previous comment — I initially fixed the two broken conditions; I've reverted to **the export verbatim** per your standing instruction, with the defects pinned by tests instead. The rules are now byte-identical to the BA site.

Adds the part of WI-001827 that was missing: **the Assignment Rules**. The work item links to three for Work Permit and none existed on this instance — only the unrelated `Work Permit Completion - GR Operator`.

| rule | Process Task created | assignee |
|---|---|---|
| `Work Permit-PRO` | Assigning PRO in Work Permit | HR-EMP-00114 |
| `Work Permit - GRD Supervisor` | Assigning GRD Supervisor in W | HR-EMP-00775 |
| `Work Permit - GR Operator` | Set new Work Permit Expiry Date and mark as completed. | HR-EMP-00114 |

Rule type is **Based on Process Task**, so the assignee comes from the task's employee — an unlinked rule assigns nobody. The patch reuses a matching task rather than inserting a second copy (re-ran it to confirm: still 4 tasks), and verifies afterwards since `create_assignment_rule` logs failures instead of raising.

### Two defects in the export, shipped as given

**1. `Work Permit - GRD Supervisor` cannot evaluate.** It tests `doc.workflow_state`, but `assign_condition` is evaluated as `frappe.safe_eval(condition, None, doc.as_dict())` — the document's own fields are the locals, so `doc` is not a name in scope. Verified: `NameError: name 'doc' is not defined`, which `AssignmentRule.safe_eval` swallows into an "Auto assignment failed" msgprint. **The rule never assigns anyone, and every Work Permit save shows that message.** Dropping the `doc.` prefix is the entire fix.

**2. No rule matches the payment state.** The conditions spell it `"Pending For Payment"`; the workflow state is `"Pending  For Payment"` — **two spaces**. Nobody is assigned while a transfer waits for payment. Adding the space is the entire fix.

Both are recorded as named tests (`test_the_supervisor_rule_cannot_evaluate_at_all`, `test_no_rule_matches_the_payment_state`) so a correction on the BA side surfaces as a failure that says exactly what changed, rather than passing silently.

### Workflow

The fixture on this branch already matched the export exactly — 14 states, all transitions, roles and conditions — with one exception: the export carries `Pending By PAM --Reject--> Rejected` **twice**. That row is now duplicated here too. The two are identical, so `apply_workflow`'s last-match-wins picks the same transition and `add_dropdown_item` drops the repeat because it keys on the label.

9 + 11 tests, applied and verified on this instance.

yusuffGoodie:
Both defects corrected — and an **end-to-end check turned up a third** that was the actual reason nobody got assigned.

### 1. `doc.workflow_state` → `workflow_state`
`assign_condition` is evaluated as `safe_eval(condition, None, doc.as_dict())` — the document's own fields are the locals, so `doc` was never a name in scope. `NameError`, swallowed into an "Auto assignment failed" msgprint on every save.

### 2. `"Pending For Payment"` → `"Pending  For Payment"`
Two spaces, as the workflow spells it. Nothing matched the one-space form.

### 3. Hand-off states moved from `close_condition` to `unassign_condition`
This one only appeared once I saved a real permit instead of calling `apply_assign` directly. `apply()` runs `close_assignments` for **every** rule, and it closes **every** assignment on the document via `assign_to.close_all_assignments` — not just the rule's own. The operator's close list named `Pending By Supervisor` and `Pending  For Payment`, the two states the *supervisor* holds, so **the operator rule closed the supervisor's ToDo in the same save that created it.** The trace showed the ToDo created and then flipped to `Closed`.

Those states are now `unassign_condition`, which fires only for the rule that owns the assignment (`apply_unassign` checks `self.name in [d.assignment_rule for d in assignments]`). `close_condition` keeps the terminal states only. This is the idiom the existing one_fm hand-off chains already use (`bonus_request_hr_manager`, `bonus_request_finance_manager`).

### Verified by walking the chain on a real permit

| state | holder |
|---|---|
| Apply Online by PRO | `Work Permit-PRO` |
| Pending By Supervisor | `Work Permit - GRD Supervisor` |
| Pending By PAM | `Work Permit - GR Operator` |
| Pending  For Payment | `Work Permit - GRD Supervisor` |
| Pending By Operator | `Work Permit - GR Operator` |
| Completed / Rejected | nobody |

14 tests, three of which save a real permit and assert who holds it. Mutation-checked: restoring any of the three corrections fails the suite with a named failure. #6647 restacked.

**For the BA:** the export needs the same three corrections applied on the BA site, or the next export will undo this.

NoBoneZ (approved):

PR #6829, opened 2026-08-31
fix: lock down Employee Checkin form fields and clean up Shift Detail…

samdanikouser:
- [x] Chore
- [ ] Feature
- [ ] Bug

Lock down the Employee Checkin form so key fields can no longer be edited directly through the UI, and clean up the Shift Details section layout.

- Made Employee, Time, Skip Auto Attendance, Location / Device ID, Shift Assignment, and Roster Type read-only.
- Removed the Fetch Geolocation button.
- Merged the Shift Timings fields (Shift Start, Shift End, Shift Actual Start, Shift Actual End) into the Shift Details section and removed the separate Shift Timings section title, so the section renders as one clean 2-column block instead of two stacked sections.
- Removed the "Check-in Form" default value from the Source field (the option itself is still selectable, it's just no longer pre-filled).
- Backfilled existing Employee Checkin records that had Source = "Check-in Form" (picked up from the old default) back to blank.

N/A

## Solution description
Employee Checkin belongs to hrms, not one_fm. Two different mechanisms are used depending on who owns each field:

- one_fm/custom/property_setter/employee_checkin.py (new) — get_employee_checkin_properties(): Property Setters for fields hrms itself owns (can't be edited directly, since that would mean editing hrms's own doctype JSON, which gets overwritten on the next hrms update).
  - read_only = 1 on employee, time, skip_auto_attendance, device_id.
  - hidden = 1 on fetch_geolocation.
  - hidden = 1 on shift_timings_section and column_break_vyyt (both explicitly kept in field_order, just hidden - omitting them entirely caused Frappe to silently re-insert them at their original core position, per frappe/model/meta.py's field_order fallback logic).
  - Reordered field_order so Shift Actual Start / Shift Actual End render inside Shift Details instead of under their own Shift Timings heading.
- one_fm/custom/custom_field/employee_checkin.py — for fields one_fm itself added (owns outright), updated the Custom Field records directly instead of via Property Setter:
  - source: cleared default: "Check-in Form" -> "".
  - shift_assignment, roster_type: added read_only: 1.
- one_fm/setup/property_setter.py — registered the new property setter getter in get_field_properties().
- one_fm/patches/v15_0/update_employee_checkin_form_customizations.py (new) — applies the Property Setters and re-syncs all Employee Checkin custom fields via create_custom_fields(get_employee_checkin_custom_fields()) on existing sites (fixture file edits alone don't touch already-installed sites; after_install only runs on brand-new sites, so a patch is the only way to push this to existing ones).
- one_fm/patches/v15_0/backfill_employee_checkin_default_source.py (new) — clears source on existing Employee Checkin records where it's currently "Check-in Form", using the Frappe Query Builder (not raw SQL), batched (5,000 rows/commit) since the table has ~2.5M rows and no index on source. Only the source column is ever set - modified/modified_by are never touched (verified against live records before/after). Only rows equal to "Check-in Form" are touched - Mobile App, Mobile Web, Frappe Page, Attendance Check, and already-blank rows are untouched.
- one_fm/patches.txt - registered both new patches.

No changes were made to one_fm/public/js/doctype_js/employee_checkin.js - an earlier attempt to also strip the geolocation map's "locate me" control was reverted per review feedback, so the client script is unchanged from version-15.

- [x] No
- [ ] Yes

Purely metadata (Property Setters / Custom Fields) and a one-time data backfill - no controller/hook logic added.

None attached.

## Areas affected and ensured
- Employee Checkin doctype form (HR module) - field read-only states, Fetch Geolocation button, Shift Details section layout, Source field default.
- Existing Employee Checkin data - source column backfilled for ~670k records.

Yes.
- Users (including System Manager, previously able to edit via frm.disable_form() bypass) can no longer manually edit Employee, Time, Skip Auto Attendance, Location/Device ID, Shift Assignment, or Roster Type from the Employee Checkin form - these are now enforced read-only at the field level, not just via the client-side disable_form() role check.
- The Fetch Geolocation button is gone from the form.
- New Employee Checkin records no longer default Source to "Check-in Form".
- Existing records that had Source = "Check-in Form" now have it blank (modified/modified_by untouched).

- [x] Existing Data
- [ ] New Data

Verified against onefm.localhost (~2.5M existing Employee Checkin records) via frappe.get_meta() introspection and direct DB queries after running both patches.

## Was child table created?
No child table involved.
- [ ] is attachment required?

- [ ] Yes
- [x] No

(Custom field attributes - default/read_only - were updated, none deleted.)

## Is patch required?
- [x] Yes
- [ ] No

Two patches: update_employee_checkin_form_customizations (Property Setters + Custom Field sync) and backfill_employee_checkin_default_source (data backfill).

### Was the patch tested?
Yes - both run against onefm.localhost:
- update_employee_checkin_form_customizations: verified via frappe.get_meta("Employee Checkin") and direct tabCustom Field / tabProperty Setter queries that all read-only/hidden flags, defaults, and field_order applied correctly.
- backfill_employee_checkin_default_source: verified via DB count before/after - 669,893 records with source = "Check-in Form" -> 0 remaining, all other source values (Mobile App, Mobile Web, Frappe Page, Attendance Check) unchanged. Confirmed modified/modified_by are untouched by manually resetting a few records and re-running the patch.

- [ ] Chrome
- [ ] Safari
- [ ] Firefox

Testing was done via backend introspection (bench console / direct DB queries), not a manual browser pass - recommend a quick visual check in the browser before merge.

Screenshot 2026-09-01 at 11 26 27 AM
Screenshot 2026-09-01 at 11 26 51 AM
Screenshot 2026-09-01 at 11 28 26 AM


one_fm/patches/v15_0/update_employee_checkin_form_customizations.py
```diff
@@ -0,0 +1,23 @@
...
+
+from one_fm.custom.property_setter.employee_checkin import get_employee_checkin_properties
+from one_fm.setup.setup import add_property_setter
+
+
+def execute():
+    add_property_setter(get_employee_checkin_properties())
+
+    custom_fields = {
+        "Employee Checkin": [
```
ks093: @samdanikouser This is not a standard field. Source is a custom field. We don't add property setter for custom field. 
We directly update the custom field.

one_fm/patches/v15_0/backfill_employee_checkin_default_source.py
```diff
@@ -0,0 +1,41 @@
...
+		# Read rowcount before commit() - commit() runs its own query on the
+		# same cursor and would otherwise overwrite it.
+		affected = frappe.db._cursor.rowcount
+		frappe.db.commit()
+
+		if not affected:
+			break
+
+		total_updated += affected
+		frappe.logger().info(
```
ks093: @samdanikouser 
1. Please remove frappe.logger statements. 
2. Use frappe query builder instead of raw sql. 
3. Make sure modified_by and last_modified is not updated when these records are updated.

ks093 (changes requested):

ks093 (approved):

PR #6931, opened 2026-09-17
WI-002539: one baseline for a leg's minutes, and two ways they stopped being lost

yusuffGoodie:
> **Stacked on [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) (WI-002538)** — base is `WI-002538-replay-a-lost-save-race`, not `staging`. Review/merge that one first; this PR's diff is only its own commit.

## The problem

Every modal that asks for a drive seeded its own number, and the server chained an untimed stop on a third:

| entry point | was | now |
|---|---|---|
| Assign Trip (solo drop) | 60 transit / 15 buffer | **15 / 5** |
| Transit to X (drop on an occupied lane) | 30 / 10 | **15 / 5** |
| Merge Into Existing Trip (drawer) | 30 / 10 | **15 / 5** |
| New Independent Trip | 60, **no buffer field at all** | **15 / 5** |
| `DEFAULT_TRANSIT_MINUTES` (server) | 30 | **15** |

So the same card timed differently depending on which way it reached the lane.

**New Independent Trip is not named in the ACs.** It creates a new trip on a vehicle lane, which is what AC1 is about, so leaving it would have kept exactly one path still timing a run on 60 minutes and no buffer. Flagging it as a judgement call rather than burying it.

## AC3 — a drive is never instant, but a blank drive still can't be confirmed

`_legTransit()` floors a leg to 1 minute **only when it has somewhere onward**. A stop at the end of a run legitimately carries 0, and inventing a minute there would move the run's end time.

This is a floor on what gets **stored**, not a default nobody chose. The `"Every drive needs its minutes before the trip can be confirmed"` guard from WI-002078 AC3.6 is **deliberately left standing** — a literal reading of AC3 would have retired it, since a transit that is never 0 makes `untimed` permanently empty and lets a pickup be scheduled on a 1-minute drive nobody typed. There's a test pinning that guard in place.

## AC4 — the actual root cause

Typing in one minute box reset the other to 0. The browser posted only the key it had just changed:

```js
timings[ship][key] = parseInt(this.value, 10) || 0;   // one key
```

…and the server refills whatever is absent from its own default — `DEFAULT_TRANSIT_MINUTES` for the drive, `0` for the buffer. So a buffer typed on a leg with no entry yet sent that leg's transit back as **0**, and the dispatcher watched the number they had just read reset itself. The handler now reads **both** inputs off the row, so the pair always travels together.

## Slices of WI-002578 that came along

Opening the Trip Builder from the Assign modal (your instruction) needed two things that belong to 578. Both were cheap:

- **The preview refused a run of one card** (`len(names) < 2`). But `build_itinerary` *already* gives a single outward card the camp → site → home shape — which is precisely the 2-leg round trip **578 AC7** describes. Relaxing the guard was the whole change; no new itinerary logic.
- **Confirm can't merge a run of one** — `merge_trip_shipments` genuinely needs two cards to mint a `trip_group` — so a solo run applies its preview directly and keeps its own direction. That also unblocks **Edit Trip Timings for single trips (578 AC6)**, which the old guard sent the operator away from.

I'll note this on WI-002578 so those two ACs aren't built twice.

Also fixed while in there: the rejected-save rollback called `undo_merge` on `merged.itinerary.map(s => s.shipment)`. For a solo run those are leg keys (`leg-2`), not shipment names, and nothing was merged anyway — the cards actually merged are now named explicitly.

## Tests

`one_fm/tests/test_default_leg_timings.py` — 18 tests across the four ACs, including one that asserts the AC3.6 blank-drive guard survives, and two that build a real single-card preview and assert it produces the 3-stop round trip.

Two tests in `test_merge_preview` pinned the old literals (`DEFAULT_TRANSIT_MINUTES == 30`, and the expression the transit floor replaced). Both still guard their original point and were **updated to the current values, not deleted**.

| module | result |
|---|---|
| `test_default_leg_timings` | 18 passed |
| `test_merge_preview` | 97 passed |
| `test_mixed_trip_merge` | 62 passed |
| `test_forward_departure_scheduling` | 48 passed |
| `test_transportation_seat_rule` | 36 passed |
| `test_transportation_shipment` | 33 passed |
| `test_transportation_schedule` | 27 passed |
| `test_transportation_manifest_page` | 26 passed |
| `test_one_trip_group_is_one_run` | 18 passed |
| `test_save_conflict_retry` | 15 passed |

**Not yet verified in a browser.** The Playwright MCP server could not connect (it needs Node ≥ 20; this bench defaults to 18) and no site was running. Both now fixed locally, so the visual half of AC1 — the renamed button and the Trip Builder opening seeded — is click-testable next session.


yusuffGoodie:
## Verified in the browser

Dragged a pool card onto an empty lane on the live board:

**Assign modal**
```
title:          Assign Outbound (→ To Site) — DHL Ardiya
Buffer Time:    5      ← AC1
Transit Time:   15     ← AC1
primary button: Open Trip Builder
```

**Clicking it opened the Trip Builder, seeded:**

| Stop | Place | Action | Buffer | Transit | Departs | Arrival |
|---|---|---|---|---|---|---|
| 1 | Mahboula Camp | Boarding | **5** | **15** | 14:40 | 15:00 |
| 2 | DHL Ardiya | Dropping Off | 5 | 15 | 15:00 | 15:20 |
| 3 | Mahboula Camp | Dropping Off | — | — | 15:20 | — |

Three things that confirms at once:

- the minutes entered in the Assign modal **flow into the first leg**, as asked;
- a **single card** produced the full camp → site → home round trip (the WI-002578 AC7 shape) with no new itinerary logic;
- stop 3 has nowhere onward and correctly shows **—**, not a floored 1 — the `_legTransit` "drives" test holds in practice.

**AC4 live:**

| action | buffer | transit |
|---|---|---|
| opened | 5 | 15 |
| set buffer 12 | **12** | **15** ← was reset to 0 before |
| set transit 27 | **12** | **27** |

Leg 2 stayed 5/15 throughout, and the walk followed: 14:40 + 12 + 27 = **15:19**.

"Edit Trip Timings" is also enabled on a single-stop run now (WI-002578 AC6).

The block this created was removed afterwards, so the plan is as it was found.

NoBoneZ (approved):

PR #6932, opened 2026-09-17
WI-002540: one stop, several stops, or the whole run comes off a lane

yusuffGoodie:
> **Stacked on [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) (WI-002539) → [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) (WI-002538).** Base is the 002539 branch, so this diff is only its own commit. Merge order: 6930 → 6931 → this.

## AC1 — the three removal actions

| action | before | now |
|---|---|---|
| Remove a single stop leg | ✅ existed | unchanged behaviour |
| Multi-select and remove several stops | ❌ | tick per stop in the drawer + **Remove N Selected Stop(s)** |
| Remove the entire trip block at once | ❌ | **Remove Entire Trip** (confirmed first) |

All three go through **one** `_removeItems(ids)` path. They differ only in which block ids they collect; what follows — a run that may now travel only one way, a card with nothing left on any lane returning to the pool, the plan being written — is identical and previously lived inside `removeSelectedFromLane` alone. Writing the two new controls against their own copy is exactly how a run ends up still flagged `MIXED` after the stop that made it mixed has gone.

Two details worth pointing at:

- **A card returns to the sidebar only when nothing of it is left anywhere.** A card placed in both directions keeps its pool entry while one leg still stands — so removing the outbound leg of a two-leg card does *not* wrongly free it.
- **The direction resync is per trip, not per `tripId`.** A multi-stop removal can empty stops from more than one run at once.

## Ticks can't leak between runs

A tick left on trip A while the dispatcher moves to trip B would otherwise be swept up by a removal made on B. The action reads a `checkedStopIds` computed scoped to the run the drawer is *showing*, which also keeps the button's count and its action in step by construction; ticks are cleared when the drawer closes and after every removal.

## The two Notes — already true, now pinned

> *Removing Stops should update Route Plan doctype instantly.*
> *Removing Stops should update Transportation Shipment doctype and change the status to Unassigned.*

Neither needed new code. `save_assignments` rewrites the plan's assignment rows from what is left, and `_sync_shipment_statuses` reverts every shipment this plan dropped back to `Unassigned` while leaving the stops beside it alone. The removal path ends in `persistAssignments`, which posts immediately — and "instantly" only became reliably true once [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) stopped those saves colliding.

Rather than re-implement them, this PR adds tests that **fail if either stops holding**, including a real `Transportation Shipment` round-trip asserting a dropped card goes to `Unassigned` and a still-placed one stays `Assigned`.

## One judgement call

AC1 says the controls may live in "the RHS details drawer **or** trip builder". I put them in the drawer — that's where selection lives and where removal already was. Adding a second removal surface inside the Trip Builder would mean re-running the server preview with fewer shipments on every tick, which is a materially bigger change. **Say the word if you want it in the Trip Builder as well.**

## Tests

`one_fm/tests/test_stop_removal_controls.py` — 16 tests.

| module | result |
|---|---|
| `test_stop_removal_controls` | 16 passed |
| `test_merge_preview` | OK |
| `test_mixed_trip_merge` | OK |
| `test_transportation_schedule` | OK |
| `test_transportation_shipment` | OK |
| `test_one_trip_group_is_one_run` | OK |

**Not yet click-tested.** Playwright's MCP server still can't attach to this session (the Node-20 fix needs a restart), so the drawer's tick UI is verified by source and unit tests only.


yusuffGoodie:
## Verified in the browser

Opened the RHS drawer on a real 3-stop run:

```
footer buttons (nothing ticked):
  Edit Trip Timings | Reassign Vehicle
  Remove Stop 1 (WTE Sabah Al Ahmed) from Lane
  Remove Entire Trip
```

| action | footer |
|---|---|
| tick 1 stop | `Remove 1 Selected Stop(s)` appears |
| tick a 2nd | count updates to `Remove 2 Selected Stop(s)` |
| untick both | the button disappears again |

Three checkboxes rendered for three stops, and **ticking never changed which stop was selected** — the single-stop button kept reading `Remove Stop 1 (WTE Sabah Al Ahmed) from Lane` throughout, so the `@click.stop` guard holds.

`Remove Entire Trip` correctly does **not** render on a single-stop run (`selectedTripStops.length > 1`) — confirmed separately on a 1-stop block, where only `Remove from Lane` was offered.

A real removal was then performed: block count went **106 → 105**, the card left the lane immediately, and **no "Document has been modified" dialog appeared** — which is also a live check on [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) AC2. Console showed no JS errors (only a pre-existing 500 on `/files/onefm1.ico`).

NoBoneZ (approved):

PR #6933, opened 2026-09-17
WI-002614: the manifest stops doing day arithmetic on the lock lifespan

yusuffGoodie:
> **Stacked:** [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) → [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) → [#6932](https://github.com/ONE-F-M/one_fm/pull/6932) → this. Base is the 002540 branch, so the diff is only its own commit.

## One field, two symptoms

A `Route Plan Assignment`'s `start_time`/`end_time` carry **two different things**: the TIME is the daily trip window, the DATE is the multi-day vehicle lock's lifespan (TR-8). They're stamped when the lock is set, not when the bus drives — so two stops of one run routinely hold unrelated dates.

The manifest was subtracting those timestamps **whole**. That one expression, in two places, is both bugs in the report:

| symptom | cause |
|---|---|
| `+11 DAY`, `+12 DAY`, `+27 DAY` on a run that finished the same afternoon (**AC2**) | `dayOffset` took the raw date difference |
| a **24h** drive between two stops twenty minutes apart (**AC3**) | `calcTransit`'s clock-gap fallback took the same difference → ~86400s → `fmtDuration` clamps past-a-day to a flat `"24h"` |

## The fix

Both readers now go through one `secondsOfDay()` helper and take the **time of day and nothing else**, in `Asia/Kuwait` — the clock every time on this page is already printed in (`fmtTime`), so a stop can't disagree with the arithmetic done about it. Reading the raw ISO string instead would take UTC hours and silently shift everything by three.

- **AC2** — a leg crosses midnight when it arrives at a clock time *earlier* than the run left at, which is the only thing these timestamps can honestly report. Capped at one day. **AC 1.6 still holds**: a leg that really does cross 00:00 still says so (22:30 → 00:20 is `+1 Day`).
- **AC3** — the fallback wraps at midnight rather than going negative, so 23:40 → 00:20 is forty minutes, not minus twenty-three hours. The 24h clamp is now unreachable.

## AC1 — deliberately not addressed here

AC1's examples (`14:00 + 15min → 14:15`, `+25 → 14:40`, `+25+40 → 15:45`) describe the forward walk that **`walk_legs` already performs server-side** — the single rule the Trip Builder, the canvas blocks, the Route Plan rows and the manifest all read.

Re-deriving stop times in the browser would be a *third* implementation of that walk, which is precisely the drift the existing code comments warn about, and which your earlier call ("keep `walk_legs`, apply the formula per leg") rules out. The arithmetic mismatches reported look like the **same corrupted dates** rather than a separate defect — the transit labels were wrong, so the sequence read wrong.

**So AC1 needs a look at the real screen once this date noise is gone.** If stop times are still inconsistent after this, that's a genuine second defect and I'll chase it — but I'm not building a competing timing model on a guess.

## Tests

`one_fm/tests/test_manifest_timestamp_math.py` — 16 tests. The JS rules are walked in Python (including the report's own 14:00/14:15/14:40/15:45 numbers), plus exhaustive sweeps asserting **no** start/end pair can ever produce a badge above 1 day or a gap reaching a full day.

| module | result |
|---|---|
| `test_manifest_timestamp_math` | 16 passed |
| `test_transportation_manifest_page` | OK |
| `test_stop_removal_controls` | OK |
| `test_merge_preview` | OK |

**Not yet click-tested** — Playwright's MCP server needs a session restart to pick up the Node-20 fix.


yusuffGoodie:
## Verified in the browser against real plan data

Ran both rules over **RP-2026-06-1754678** — 105 runs, 597 assignment rows:

| | rows badged | worst badge |
|---|---|---|
| **old rule** (whole-timestamp subtraction) | **53** | **+39 Day** |
| **new rule** (time of day only) | **2** | **+1 Day** |

The 53 include the exact symptoms in the report:

| run | start | stop | old | new |
|---|---|---|---|---|
| MIX-258399717d82 | 2026-08-18 11:18 | 2026-08-29 11:52 | **+11 Day** | 0 |
| MIX-258399717d82 | 2026-08-18 11:18 | 2026-08-30 12:07 | **+12 Day** | 0 |
| MIX-a762fdd92886 | 2026-08-18 00:55 | 2026-08-30 02:30 | **+12 Day** | 0 |

Note the times in each pair: 11:18 → 11:52 is **34 minutes apart**. Only the DATE differs — and that date is the vehicle-lock lifespan, never the run's day. That is the whole bug.

The 2 surviving badges are genuine midnight crossings, so **AC 1.6 still holds**.

On the live manifest page after the fix: no `+N Day` badge anywhere, and no `24h` transit.

Still open as stated in the description: **AC1** wants checking against the real screen now the date noise is gone.

NoBoneZ (approved):

PR #6934, opened 2026-09-17
WI-002590: each run on a vehicle checks in on its own

yusuffGoodie:
> **Stacked:** [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) → [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) → [#6932](https://github.com/ONE-F-M/one_fm/pull/6932) → [#6933](https://github.com/ONE-F-M/one_fm/pull/6933) → this. The diff is only its own commit.
>
> ⚠️ **Adds a DocType field — needs `bench migrate` before this branch runs.**

## AC1 — the leak

A `Transportation Manifest` is one **vehicle** for one **day**, and a vehicle drives several runs in a day. The attendance-check pointer was a single `Int` on the manifest, so every run read the same number:

| action on S-801 | what happened to S-802 |
|---|---|
| Trigger Attendance Check | its controls changed too |
| Complete & Lock Stop 1 | its stop 1 was frozen as well |
| moving past stop 1 | `enforce_stop_locking` froze S-802's stop 1, because the numbers matched |

The pointer now lives per run in `active_stop_by_trip`, and every reader takes the run's own value — `_build_sheet`, `enforce_stop_locking`, `save_stop_checks`, the trigger/complete endpoints, the schedule page's `vehicleMeta`, and the manifest page.

## The part that touches verified data

A manifest that is **already half-checked when this ships** carries one number for the whole vehicle. Starting every run at 0 would silently **reopen stops a supervisor had verified** — the one thing this lock exists to prevent.

So on first read, every run **inherits** the old vehicle-wide number. Nothing that was locked comes unlocked. Once a run's pointer has been written, the map is authoritative and the runs move independently. The flat `active_stop_sequence` is kept in step as the furthest any run has reached — so anything still reading it sees something sane — but **no lock is decided from it any more**.

Three tests cover exactly this migration path, including the "never started checks" and "split map wins once it exists" cases.

`save_stop_checks` additionally rejects a row from a *different* run that numbers its stops the same way: stop 1 of S-802 is not stop 1 of S-801.

## AC2 — the buttons name their run

`Trigger Attendance Check — S-801` and `Complete & Lock Stop 1 (S-801)`. Two identical trigger buttons on one vehicle page is an invitation to check in the wrong run.

## AC3 — no trigger where nobody boards

A return-only run leaving its camp empty (`EMPLOYEES BOARDING: 0`) used to offer a check over an empty list, and running it advanced that run's pointer past a stop with nothing to verify. The suppression is on the **trigger only** — a supervisor can still close a stop they have already opened.

## Tests

`one_fm/tests/test_trip_isolated_attendance.py` — 19 tests, including real saves asserting that a completed stop on S-801 throws while the *same stop number* on S-802 stays editable.

| module | result |
|---|---|
| `test_trip_isolated_attendance` | 19 passed |
| `test_transportation_manifest_page` | OK |
| `test_manifest_timestamp_math` | OK |
| `test_transportation_manifest` | 5 pre-existing failures |
| `test_manifest_compiler` | 1 pre-existing failure |

Those 6 failures were **re-baselined on a stashed clean tree and are byte-identical** — I did not rely on the counts matching, because `enforce_stop_locking` is precisely what this PR rewrites. All six are reliever/roster assertions untouched by this work.

**Not click-tested** — Playwright's MCP server attaches only at Claude Code start-up, so the Node-20 fix needs a restart of the editor session, not of bench.


yusuffGoodie:
## Verified in the browser

The live manifest page renders **seven runs on one vehicle**, each with its own named trigger:

```
Trigger Attendance Check — S-401
Trigger Attendance Check — S-402
Trigger Attendance Check — S-403
Trigger Attendance Check — S-404
Trigger Attendance Check — S-405
Trigger Attendance Check — S-406
Trigger Attendance Check — S-407
```

That is **AC1 and AC2 in one screenshot**, and it shows how real the leak was: seven runs were sharing a single `active_stop_sequence`, so triggering any one of them moved the pointer the other six were reading.

`get_manifest_data_for_plan` returns 200 with the new `active_stop_by_trip` on `vehicleMeta`, and the page renders clean — no JS errors beyond a pre-existing 500 on `/files/onefm1.ico`.

Verified after `reload_doctype`; a full `bench migrate` is still needed on any other site.

yusuffGoodie:
## Fix: every run can start its own attendance check

Reported by @y-oyedele: trips across several vehicles have **no Trigger Attendance Check button**, and they are not return-only runs.

### Six of seven runs on one vehicle

The rule read *"the first DEPART card of the run"* as `seq === 1`. But `seq` is the **rider's stop number across the whole run**, not an ordinal for the camp they board at — so only a trip whose boarders happen to start at stop 1 ever matched.

On **VHL-L-0004**:

| trip | S-101 | S-102 | S-103 | S-104 | S-105 | S-106 | S-107 |
|---|---|---|---|---|---|---|---|
| `camp.seq` | **1** | 3 | 5 | 8 | 9 | 3 | 13 |
| trigger button | ✓ | ✗ | ✗ | ✗ | ✗ | ✗ | ✗ |

Six of the seven runs could not start a check at all — which is most of what this work item is for. S-102's DEPART card reads **STOP 3**, and that 3 is what the rule was comparing against 1.

### What changed

The intent is unchanged and still holds: within a run only the **first pickup** may start a check, so a driver cannot begin at a mid-route camp they have not reached (WI-002074). It is now decided by the camp's **position in the run**, which is what "first" meant all along.

An ordinary camp-by-camp run still walks in turn — the stored lock is a `seq`, so its position in that run's own camp list is found and the next one along is offered.

**`seq` remains what the trigger sends and what the lock state is read back against**, so the bookkeeping either side of this is untouched.

### Verification

**30 tests**, 11 new. The decision is **executed under node**, not grepped — which camp may start a check is a position, and `seq` is exactly the value that was mistaken for one:

| case | result |
|---|---|
| run starting at stop 1 | can trigger |
| **run starting at stop 3** | **can trigger** |
| mid-route camp | cannot (WI-002074 preserved) |
| run already triggered | not offered again |
| ordinary run, first camp | can trigger |
| ordinary run, next after active | can trigger |
| ordinary run, camp already passed | not offered again |
| lock whose seq matches no camp | offers nobody — a driver must not restart a check already running |

One WI-002614 test pinned the old `seq === 1` string; its assertion moved here with the rule, so each branch stays consistent with its own code.

Full epic green on the rebased tip: timestamp math 39, trip duration 38, merge preview 97, manifest page 26, route plan 96, and the rest.

> Stack rebased: #6935–#6941 force-pushed.

yusuffGoodie:
## Fix: a run's second camp can mark attendance too

Follow-up to the previous comment, raised by @y-oyedele on **S-401**: when a trip departs from two or more camps the Trigger button appears only at the first, and the second camp's employees can never mark attendance. **Six people board at Mangaf 182 on that run.**

### It was not only the button — the API refused it as well

Both sides read `stop_sequence` as an ordinal *within a run*. It is numbered across the **vehicle**:

```
S-401 -> 1 and 5      S-402 -> 2      S-403 -> 3
S-404 -> 4            S-405 -> 1 and 5
```

So:

- the server's `stop_sequence == active + 1` could never match — completing stop 1 left the pointer at **2**, and stop **5** failed the contiguity test;
- the page pinned a merged run to its first camp and nowhere else, on the reading that a merged run leaves *one* origin. It can load at two — the same assumption WI-002614 had to correct in the itinerary — and the second camp's card then stayed **"Locked until triggered"** all day.

### Both now walk the run's own stops

Read off the manifest's rows, stepping over the gaps that belong to other runs on the same vehicle. Verified against the live manifest:

```
S-401  active=0 -> trigger 1
       active=1 -> REFUSED (stop 1 still open)
       active=2 -> trigger 5          <- previously unreachable
       active=5 -> REFUSED (stop 5 still open)
       active=6 -> none left
S-402 sequences: [2]    S-403 sequences: [3]    (unchanged)
```

**WI-002074's concern is intact** and is most of what the tests check: a camp is offered only once the one before it is **complete**, never merely triggered, so a driver still cannot start a check at a pickup they have not reached. A stop left open is refused with *"complete it before starting another"* rather than silently allowing a second.

### Verification

**34 attendance tests**, 10 new, half of them the server walk. The page decision is **executed under node** across the whole lifecycle — first camp, second camp before and after completion, an open camp, a completed one, and nothing left — because which camp may start a check is a position, and `stop_sequence` is exactly the value that was mistaken for one.

Green on the rebased tip: timestamp math 39, trip duration 38, merge preview 97, drop-time 21, manifest page 26.

> Stack rebased: #6935–#6941 force-pushed; all nine branches verified in sync.

NoBoneZ (approved):

PR #6935, opened 2026-09-17
WI-002591: the bus carries who is travelling, not who the master says

yusuffGoodie:
> **Stacked:** [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) → [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) → [#6932](https://github.com/ONE-F-M/one_fm/pull/6932) → [#6933](https://github.com/ONE-F-M/one_fm/pull/6933) → [#6934](https://github.com/ONE-F-M/one_fm/pull/6934) → this.
>
> ⚠️ **This is AC5 only.** AC1–AC4 are deliberately not here — see "What I found" below. Please read that section before reviewing.

## What I found, as you asked

You said to check the daily job and the Generate Shipments button before changing anything. Three things:

**1. The "on-demand fallback" AC5 asks for largely already exists.** `generate_transportation_shipments` *is* the nightly job, and the canvas button has always called the same whitelisted function — its own docstring says "Entry point for both the daily scheduler and the canvas Generate button". So AC5 isn't asking for a new mechanism.

**2. What was actually missing is this branch:**

```python
elif existing.status == "Unassigned":
    ...rewrite...
# Assigned → leave untouched      ← AC5's real gap
```

A crew change after a card was placed never reached the driver's manifest. The only way to pick it up was to unassign the card and re-plan the run — precisely the manual work AC5 wants removed. **That is what this PR fixes.**

**3. Two blockers for AC1–AC4** (detail below).

## The fix

`_refresh_assigned_roster` moves the roster and the headcount and **nothing else**:

- `generation_key` / `pair_group` are what the Route Plan Assignment row points at — rewriting them would orphan the very placement this preserves. Pinned by a test.
- `status` stays `Assigned`; the run keeps its lane, vehicle and trip.
- A card whose crew has **not** changed is not saved at all. Otherwise the nightly job would bump `modified` on every placed card each morning and the canvas would spend the day losing races with it ([#6930](https://github.com/ONE-F-M/one_fm/pull/6930)).

The button reports re-crewed runs separately — "updated" for both reads as though the plan had been touched.

The child-row mapping was extracted to `_write_roster` so the two writers can't drift.

## Why AC1–AC4 are not here

**Blocker 1 — "the selected date" does not exist on this screen.** `get_grouped_employees_by_accommodation()` takes **no date parameter at all**. It reads `Employee.shift` — the *master* allocation — so it always builds "today's master data", never a chosen date. The Route Plan is a standing plan (`effective_from`/`effective_until`) holding standing cards. AC1 ("for any date within that range"), AC3 ("the subsequent date") and AC5's "for the selected date" all assume a date the canvas has no concept of. Adding one is a real design change, not a fix.

**Blocker 2 — the leave states AC1 names have no rows.** AC1 keys on "marked on Leave (Annual/Sick) in the HR roster". Actual `Employee Schedule.employee_availability` counts on this DB:

| value | rows |
|---|---|
| Working | 2,102,482 |
| Day Off | 175,272 |
| Client Day Off | 11,664 |
| Client Event | 3,276 |
| On-the-job Training | 1,766 |
| … | … |
| **Annual Leave** | **0** |
| **Sick Leave** | **0** |

Keying AC1 on those two values would match nothing.

**The good news:** the reliever model already exists and is populated — `Employee Schedule.is_relieving_schedule` (**2,329 rows**) with `relieving_employee_schedule` pointing at the covered person's schedule. Real example: `2026-09-10_HR-EMP-04267` (Working, relieving) → `2026-09-10_HR-EMP-03343` (Day Off). So AC1/AC2 are buildable on the *actual* mechanism — relief against a **Day Off**, not Annual/Sick — which is also the real-world majority. Note `is_replaced` is only set on 40 rows, so the back-link is unreliable; the dependable direction is reliever → original.

**So I need two decisions before AC1–AC4:** (a) does the Transportation Schedule get a date selector, or should the overlay resolve "today" only? (b) is relief-against-Day-Off the intended trigger, given Annual/Sick Leave never appear here?

## Tests

`one_fm/tests/test_assigned_card_recrew.py` — 10 tests, including real saves asserting the crew changes while `generation_key`/`pair_group`/`status` do not, and that an unchanged crew leaves `modified` alone.

| module | result |
|---|---|
| `test_assigned_card_recrew` | 10 passed |
| `test_transportation_shipment` | OK |
| `test_shipment_deduplication` | OK |
| `test_driver_exclusion` | OK |
| `test_transportation_schedule` | OK |


yusuffGoodie:
## AC1–AC4 added (second commit) — and your description confirmed

You said leave approval deletes the Employee Schedule, creates no Shift Assignment, and marks Attendance `On Leave`. **Confirmed in both the code and the data:**

| claim | code | live data |
|---|---|---|
| ES deleted | `LeaveApplicationOverride.clear_employee_schedules` | **0** ES rows across every sampled leave range |
| no shift assignment | `close_shifts` | **0** SA rows in those ranges |
| attendance `On Leave` | `update_attendance` | 13/13, 19/19, 17/17 days; **69,427** rows total |

That corrects what I wrote in the original description above. Annual/Sick Leave has zero `employee_availability` rows **because the row is deleted entirely**, not because leave is recorded elsewhere — so the only place the date *range* survives is the Leave Application. That is now what the overlay reads.

**The bug, with a number:** an employee on leave stays `Active` with `shift` intact, so nothing removed them. **Six employees were on approved leave and still carried a shift** on the day this was written — still being placed on transportation cards while away.

## What the overlay does

`roster_overlay.py`, resolved per date, **stored nowhere**:

- **AC1** — drop anyone on approved leave; add anyone relieving, filed under the shift they are **covering** rather than their own.
- **AC2** — who a rider stands in for is stamped on the card row, and the drawer's existing Reliever badge gains the tooltip. Stamped rather than resolved at render time: the manifest is printed and carried, and must still make sense tomorrow when the schedule behind it has moved on.
- **AC3** — free. Nothing is stored, so a range ending restores the original rider by itself. No expiry job.
- **AC4** — exactly as you described: the map is built from the master allocation, so a shift change re-files the employee; with the AC5 re-crew that now reaches **assigned** cards too.

## Two things for the reviewer

**1. No reliever in the database covers someone on leave — 0 of 2,329.** They all cover a **Day Off**, precisely because the leave flow deletes the schedule a reliever would point at. So leave-removal and reliever-substitution both work, but as two *independent* effects rather than the single chain AC1 describes. Flagging rather than forcing the data to fit the wording.

**2. `is_reliever` already meant something else.** It was a standing Employee property — "rambo reliever by trade". The new row flag means "standing in for somebody on this card today". Both are kept and OR'd, so nothing relying on the old meaning changes, and a *regular* employee covering a colleague is now badged at all.

## Tests

`test_roster_overlay.py` — 17 tests: leave boundaries, unapproved applications ignored, dangling back-links still boarding, tooltip span present/absent, and the shift map rewrite.

Two fixture traps worth noting for anyone writing similar tests: `FrappeTestCase` rolls back per **class**, not per test, so each test takes its own date; and the roster is generated **years ahead** (every date in 2033 already carries rows), so "a far-future date" is not an empty one — the fixtures pick employees who are free on the chosen day.

Two of my own AC5 source-grep assertions broke when the reliever context was threaded through. They now match the call rather than its argument list, so a later parameter cannot fail them without a real regression.

| module | result |
|---|---|
| `test_roster_overlay` | 17 passed |
| `test_assigned_card_recrew` | 10 passed |
| `test_transportation_shipment` / `test_transportation_schedule` / `test_default_leg_timings` / `test_merge_preview` | OK |

⚠️ Adds fields to **Transportation Shipment Employee** — needs `bench migrate`.

yusuffGoodie:
## Added: a prune guard (found during Operations-023 testing)

Folding a **pre-existing** defect fix into this PR, since it already owns `shipment_generator.py`. It is not a regression from this stack — I checked.

**What happened on the test site:** one click on **Generate Shipments** deleted **551 unassigned cards**, and a second click could not bring them back.

```
get_grouped_employees_by_accommodation()  ->  {}
current_keys                              ->  set()
_prune_stale(set())                       ->  "not in current_keys" is true for EVERY card
```

The demand builder returns `{}` for several unremarkable reasons — no Active Operations Shift, nobody allocated to one, or no `Accommodation Checkin Checkout` carrying an employee. The last is what bit us: the site has 120 `IN` rows and **all 120 belong to external contract tenants with no employee link**. The builder logs that as a degraded state rather than raising, so what reached the caller was an empty dict indistinguishable from *"nobody needs a bus today"*.

**The fix:** the prune runs only when the run actually knows what the demand is. An empty pool is a conclusion to be reached from data, never from the absence of it. The summary now carries `pruned`, and the canvas warns in orange for 12s instead of reporting `0 created, 0 updated, 0 removed` in green — which is exactly what the dispatcher saw *after* the pool had already gone.

**The guard is at the call site, deliberately.** Inside `_prune_stale` it would also catch `remove_unassigned_shipments_for_trip_request`, which passes an empty key set **on purpose** to clear one trip request's cards. There's a test pinning that.

### Why this isn't from this stack

| checked | result |
|---|---|
| `_prune_stale` + its unguarded call on `staging` | present, unchanged |
| `apply_to_shift_map` (this PR's overlay) | 1,670 employees in → 1,670 out, **dropped 0** |
| stage that actually returns empty | the accommodation check-in query, untouched by these 12 PRs |

My only prior edit inside the prune was `_release_manifest_references`, which made deletion *safer*.

### Verification

**12 new tests**; all 343 tests across the Transportation epic re-run green on the rebased tip.

One note on those tests: they hand `_prune_stale` every key in the pool **minus** the seeded one. It sweeps the whole pool, so a bare `{"one key"}` would mark every real card stale — and a test that destroys the pool to prove the pool is safe is one commit away from being the bug it tests for.

The 551 deleted cards were recovered from `Deleted Document` (550 restored + 1 already present, 0 failures, 1,763 roster rows intact, the 378 Assigned cards untouched).

> Stack rebased: #6936–#6941 sit on top of this and were force-pushed. All seven remain `mergeable: true`.

yusuffGoodie:
## Second fix: take the relieved employee OFF the bus

Found during testing — thanks @y-oyedele for spotting it.

`apply_to_shift_map` added the reliever but **never removed the person they are standing in for**, so the seat was counted twice: the reliever rode from their own camp while the covered employee stayed on the original card.

On the restored production data that is **16 people travelling on their day off**, and nothing else would have removed them. Leave does not catch it — relief here is almost always against a **Day Off**, because the leave flow deletes the very schedule a reliever would point at, so there is no Leave Application behind it.

This is what AC1's *"replace the leave/absent employee's name with the assigned Reliever employee's name"* asks for.

### Verified end to end on real data

```
employees on the bus:        1,684 -> 1,668   (16 removed)
Thomas Otieno Owino          off the map      (Day Off, relieved)
Bina Thapa Magar             Security-Center point Fahaheel-Day-1
  (her own master shift is   Security-BSK - 360 Mall-Day-1)
covered employees left in:   none
```

### What this deliberately does NOT do

The criterion reads as though the reliever's name replaces the absent one **in place**, and it does when the two share a camp — **5 of today's 16**. For the other **11** the reliever sleeps somewhere else:

| | camp |
|---|---|
| Thomas (covered) | ACC-08 — **Mangaf 182** |
| Bina (reliever) | ACC-03 — **Mahboula 13** |

S-1603 and S-805 serve Mangaf 182. A same-card swap there would put a passenger on a vehicle that **cannot pick her up**. The honest result is a rider removed from one camp's card and a new card raised from the reliever's own — which is exactly what has to happen for either of them to travel, and what the generator already does (TS-1081/TS-1082).

Flagging it here because the AC's wording implies otherwise, and the process owner may want to say so explicitly.

A dangling back-link names nobody as covered, so nobody is taken off and the reliever still boards — an extra seat beats a missing passenger.

### Also corrected in this round

The RELIEVER tag itself was **already working** — the roster row carries `is_reliever: 1`, `relieving_employee_name`, `absence_reason`, the API sends them and the drawer renders them. It only renders in the **RHS drawer**, not on the unassigned sidebar card, so it is invisible until the card is placed on a lane. Worth knowing when testing AC2.

The tooltip reads `Relieving <name> | Day Off` rather than `Leave Span: … → …` whenever there is no Leave Application behind the relief — which today is all 16. Kept as-is by decision; raised with the BA separately since AC2 assumes leave-based relief that this system's data never actually produces.

### Verification

**21 overlay tests** (5 new), and all **343 tests** across the Transportation epic re-run green on the rebased tip.

> Stack rebased again: #6936–#6941 force-pushed. All seven remain `mergeable: true`.

NoBoneZ (approved):

PR #6936, opened 2026-09-17
WI-002578: a card is timed from where it was dropped

yusuffGoodie:
> **Stacked:** [#6930](https://github.com/ONE-F-M/one_fm/pull/6930) → [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) → [#6932](https://github.com/ONE-F-M/one_fm/pull/6932) → [#6933](https://github.com/ONE-F-M/one_fm/pull/6933) → [#6934](https://github.com/ONE-F-M/one_fm/pull/6934) → [#6935](https://github.com/ONE-F-M/one_fm/pull/6935) → this.

## The gap

The drop event has always carried the position on the lane, and the board threw it away — `onLaneDrop` read only the vehicle. A card released at 14:00 opened its trip on a departure re-derived from the shift window: **the operator picked a slot and the board ignored it.**

`dropTimeFrom` turns the release point into a time:
- measured against the **lane's own box**, not `offsetX` — which is relative to whichever grid line or wash rect the pointer happened to be over, so the same drop read differently depending on what it landed on;
- clamped to the lane;
- scaled from CSS pixels into the `svgWidth` space that `xToTime` speaks (they differ once the board is zoomed).

## The part that's easy to get half-right

That time has to win in **two** places:

| | what it drives |
|---|---|
| `runStart` | the clock the Trip Builder's Initial Departure opens on (**AC1**) |
| `runStartMs` | what `_applyMerge` anchors the re-timed blocks to (**AC4**) |

Setting only the first showed the dropped time in the field and then left the block sitting where the shift window put it. Both now derive from one value, and a run reopened through *Edit Trip Timings* passes no drop and still opens on its stored departure.

## Criteria already delivered — pinned, not rebuilt

- **AC2** — the forward walk is `walk_legs` on the server, the single rule the Trip Builder, canvas, Route Plan and manifest share. Per your call, not reimplemented in the browser. Verified live: 14:40 + 12 buffer + 27 transit = **15:19**.
- **AC3 / AC7** — the round trip falls out of `build_itinerary` for a single card; **AC6** unblocked *Edit Trip Timings*. Both landed in [#6931](https://github.com/ONE-F-M/one_fm/pull/6931) and are browser-verified there (camp → site → camp, last stop carrying no onward transit).
- **AC5** — the drawer already prints Trip Start, Trip End and the total above the stop sequence. Left alone.

**AC1's "bypass the assignment popup" is deliberately not implemented** — you asked for the Assign modal to stay and lead into the Trip Builder, which is what shipped in #6931. Confirmed with you before starting this.

## Tests

`one_fm/tests/test_drop_time_departure.py` — 15 tests, including ones that fail if the anchor and the field are ever decoupled again.

| module | result |
|---|---|
| `test_drop_time_departure` | 15 passed |
| `test_default_leg_timings` / `test_merge_preview` / `test_stop_removal_controls` | OK |
| `test_transportation_schedule` / `test_roster_overlay` / `test_assigned_card_recrew` | OK |
| `test_forward_departure_scheduling` / `test_lane_day_rebase` | OK |


yusuffGoodie:
## ⚠️ Deviation from AC1 — needs BA sign-off

Reported in testing by @y-oyedele: a **17:00 card dropped on a lane opened the Trip Builder on `12:34:12`**, and the itinerary walked forward from there.

That was **AC1 working exactly as written** — the release position seeded the Initial Departure Time. In use it reads as a fault, so it has been changed. Flagging it here because the AC asks for the opposite.

### Why the release position is the wrong source

A 17:00 card has to **reach** site by 17:00, so the bus leaves the camp at **16:40** — the arrival anchor less buffer and transit. Releasing the card anywhere else offered a departure with no operational meaning, and the dispatcher had to drop pixel-accurately or retype the time.

### It was also only half applied

`_doPlace` draws the block from the shift window and **never saw the release point**. So:

| | |
|---|---|
| Trip Builder said | **12:34** |
| The block on the lane said | **16:40** |

They agreed only once **Confirm & Apply** re-timed the blocks. Cancelling left them disagreeing for good.

Reproduced against live data — **TS-1090**, Outward, `start_time 17:00`, block persisted as `2026-09-20T13:40:00Z` (16:40 Kuwait):

| `current_departure` sent | departure returned | legs |
|---|---|---|
| *(none)* | **16:40:00** | 16:40 → 17:00 → 17:20 |
| `12:34:12` | 12:34:12 | **12:34 → 12:54 → 13:14** ← the reported screenshot, exactly |

### What changed

The release point now decides **which lane** and nothing else. The departure comes from the same place the block does, so the two can no longer disagree, and the field is still there for a dispatcher who wants a different time.

`dropTimeFrom` and the whole `droppedAt` thread are removed rather than left dangling — one path still passing it would put the old behaviour back on that path alone, and the two would disagree again.

**The rest of WI-002578 is untouched**: the forward walk (AC2), the automatic 2-leg round trip (AC3/AC7), block anchoring (AC4), the drawer's trip window (AC5), Edit Trip Timings on a single-stop trip (AC6), and the assignment modal handing straight over via **Open Trip Builder** (the part of AC1 that stands).

### Verification

**16 tests.** The departure rule is now **executed** against a seeded card rather than grepped:

- a 17:00 card leaves at **16:40**
- it reaches site **exactly on the shift start**
- longer legs back off further (30 transit + 15 buffer → **16:15**)
- a **typed departure still wins**
- reopening on a placed run returns the same minute, so Confirm moves nothing
- no `dropTimeFrom` / `droppedAt` survives anywhere in the canvas

All **343 tests** across the Transportation epic re-run green on the rebased tip.

> Stack rebased: #6937–#6941 force-pushed. All remain `mergeable: true`.

yusuffGoodie:
## Fix: Confirm & Apply now works on a run of one card

Reported by @y-oyedele while testing the +1 DAY badge: a leg changed to 85 minutes previewed as arriving **00:30**, Confirm closed the modal, and **nothing happened** — the block kept its width and the plan kept the old 15.

### One line

A run of **one card** never goes through `merge_trip_shipments` — that endpoint needs two cards to mint a trip group. So the modal applies the preview directly and builds `merged` **by hand**:

```js
{ trip_group: soloTripId, trip_direction: ..., merged_shipments: [] }
```

No `itinerary`. And `_applyMerge` opened by dereferencing one anyway:

```js
const order = merged.itinerary.map((s) => s.shipment);   // TypeError
```

It threw **after `d.hide()` had already run**, inside the dialog's own handler. The modal closed, the rest of the method never executed, and the block sat unmoved on its old minutes while the preview had just shown the operator the new ones. **Nothing reported a failure because nothing was left to report it** — which is exactly why it read as "Confirm does nothing".

`order` only positions a card being *merged in*, which a solo run has none of, so `[]` is the honest value rather than a guard bolted on. Guarded at the consumer, since both callers reach that line.

### Why this took three passes to find

Worth recording, because each earlier reading was wrong in a way that looked right:

| I said | Actually |
|---|---|
| the camp leg persists 0/0 | true, but a symptom — the save never ran |
| the home leg's minutes are dropped | wrong, `_applyMerge` does store `home` |
| `leg_timings` isn't persisted | true but irrelevant — it's rebuilt from the rows |

The preview was right, the server was right, and the stop→block mapping was right. Everything downstream of the throw simply never ran. The thing that finally pinned it was @y-oyedele confirming **the block does not change width** — that ruled out persistence and put the fault before it.

### Verification

**21 tests**, 5 new. The shipped line is **executed under node** against both shapes — the solo object that used to throw, and a server itinerary that must still be read — because a one-line dereference is precisely what a source grep waves through.

Also pinned: the solo branch still passes `previewStops, departureShiftMs, runStartMs` through, since dropping those would make Confirm a no-op again, quietly this time.

Full epic green on the rebased tip: 380 tests across 15 suites, plus 96 route plan and 26 manifest page.

> Stack rebased: #6937–#6941 force-pushed.

NoBoneZ (approved):

PR #6938, opened 2026-09-17
WI-002542: reorder a run's stops without a long vertical drag

yusuffGoodie:
> **Stacked:** 6930 → 6931 → 6932 → 6933 → 6934 → 6935 → 6936 → [#6937](https://github.com/ONE-F-M/one_fm/pull/6937) → this.

## A bug this found

Extracting the reorder so the arrows could share it surfaced a defect in code that was already there:

```js
if (sourceIndex >= tripStops.length || targetIndex >= tripStops.length) return;
```

`targetIndex` is an **insert-before** position, so `length` is legal — it means *"put it at the end"*. Dragging never produces that, because every drop target is an existing row, so the guard had never been wrong in production. But **the down arrow on the second-to-last stop is exactly the move that has to append**, and it was silently swallowed: the click did nothing at all, no error, no alert.

I only caught it by clicking it on the live board. The unit tests I was about to write would have asserted the arithmetic I *intended* and passed.

## The three additions

| AC | |
|---|---|
| **1** | ▲/▼ per stop; hidden at the ends of the run, so the only arrows on screen are ones that do something |
| **2** | Compact / Detailed toggle, collapsing everything below the stop header |
| **3** | Drawer auto-scrolls within 15% of either edge while dragging |
| **4** | Arrows go through the **same** `_reorderStop` as the drag |

**AC4 is why the refactor was worth it.** It asks that an arrow move re-index and re-time "exactly as before" — the only way to mean that literally is for it to *be* the same code. Three triggers, one path; a second implementation drifts the moment either is touched.

## Verified on the live board

- arrow visibility across a 3-stop run: **`[1, 2, 1]`** — first hides ▲, middle has both, last hides ▼
- moved "Wafra Substation WJO" from position 3 → 2 (alert confirmed the new order), then back down → **run restored to its original order**
- compact view: **474px → 66px** per row, chips and detail rows gone, **8 rows fit** the 531px drawer against AC2's "6–8"; toggling back restores exactly

## Two details worth a look

- **Compact collapses the row's own padding too.** Hiding the content but keeping the padding leaves the run as tall as it was, which defeats the point.
- **Auto-scroll runs on a timer, not off `dragover` alone.** `dragover` only fires while the pointer *moves*; a dispatcher holding still at the edge still expects the list to keep coming. It's cleared on drop, on drag end, and as soon as the pointer leaves the edge — a timer left running would scroll the drawer for the rest of the session.

## Tests

`one_fm/tests/test_stop_sequence_controls.py` — 24 tests. The index arithmetic is walked in Python, **including the append case that was broken**, so that regression is now pinned.

| module | result |
|---|---|
| `test_stop_sequence_controls` | 24 passed |
| `test_stop_removal_controls` / `test_pool_filters` / `test_transportation_schedule` | OK |
| `test_default_leg_timings` / `test_drop_time_departure` / `test_merge_preview` / `test_forward_departure_scheduling` | OK |


yusuffGoodie:
## Over-stretch fix (second commit)

Thanks for catching this — reproduced on **S-101** and it turned out to be a real defect that predates this PR.

### What it was

The camp departure and the ride home are stored against the **trip**, not on any block. The reorder re-timed the stops and never touched them. The lane draws a run from `min(stored departure, first stop)` → `max(stored arrival, last stop)`, so:

| S-101 on the live plan | |
|---|---|
| stops after reorder | **17:33 – 19:01** |
| stored pair (untouched) | 04:50 / 05:59 |
| block drawn | **04:50 – 19:01** ← the over-stretch |
| drawer header | `04:50 → 05:59 (**-1371 min**)` |

**This was not introduced by the arrows.** Drag-to-reorder never moved the pair either — the arrows just made it easy to reach.

### The fix, in two halves

**1. The reorder moves both stamps**, each by the delta of the end it belongs to: the camp departure keeps its lead over the first stop, the ride home keeps its trail after the last. One shared delta would put the arrival wrong whenever the reorder changes the run's overall length — which it does whenever the gaps aren't all equal.

**2. That stops the drift but doesn't repair a plan already holding a stale pair**, so the reads now refuse one that cannot describe its run: *a bus leaves before its first stop and gets home after its last*. A pair failing either test came from a different timing, and **both halves are dropped together** — keeping the plausible half still draws the block from a time the run doesn't have.

The clamp itself is unchanged, so a **real** camp leg still widens the block — that's the point of it. One rule, read by both the lane and the drawer, so the header and the block can't disagree again.

### Verified on the live board

- S-101 header now reads **`17:33 → 19:01 (88 min)`**, block draws **17:33–19:01**
- a reorder moves the stops and leaves the span correct; moving back restores the original order
- across **all 95 trips** on the plan, **none spans more than ten hours** — so no other run is still stretched

### One test updated, not dropped

`test_lane_day_rebase` pinned the old expression by its literal text. The behaviour it guards — *a stored timing may widen a block but never shrink it below its stops* — is unchanged, so the assertion moved to the current form and gained a check that both readers share one rule.

10 new tests cover the shift and the staleness rule, including S-101's actual numbers. Full sweep green: `test_lane_day_rebase`, `test_stop_sequence_controls` (34), `test_merge_preview`, `test_transportation_schedule`, `test_drop_time_departure`, `test_pool_filters`, `test_forward_departure_scheduling`, `test_mixed_trip_merge`, `test_default_leg_timings`, `test_stop_removal_controls`.

NoBoneZ (approved):

PR #6939, opened 2026-09-17
WI-002543: the Trip Builder's leg table stops scrolling sideways

yusuffGoodie:
> **Stacked:** 6930 → 6931 → 6932 → 6933 → 6934 → 6935 → 6936 → 6937 → [#6938](https://github.com/ONE-F-M/one_fm/pull/6938) → this.

## Measured, at four widths

Real run, all legs rendered:

| viewport | dialog | frame scrolls horizontally | last column | Accommodation col |
|---|---|---|---|---|
| 1854px | **95%** | **no** | flush at edge | 377px |
| 1280px | **95%** | **no** | flush at edge | 195px |
| 1024px | **95%** | **no** | flush at edge | 114px |
| 900px | **95%** | **no** | flush at edge | 75px |

The location columns absorb the squeeze — that's AC3's *"allocate remaining space proportionally to text columns"* — and `Target Arrival` / `On Board` stay fully visible at every width (AC4).

## Three changes, and why one alone wouldn't do

**1. The dialog takes 95% of the viewport.** The Trip Builder is a working surface, not a message box.

**2. `table-layout: fixed`.** Without it the browser sizes columns from their **content**, so one long location name widens the whole table however much room the dialog has — the widths would be suggestions rather than rules.

**3. `table-responsive` is gone from this table.** Its `overflow-x: auto` **is** the sideways scrollbar AC1 asks to eliminate. Widening the dialog while leaving it in place only moves the problem. The frame now scrolls *down* and never *across*, with the header sticky while it does.

## Details worth a look

- **Every fixed width is a number or a clock** — Stop 44, Action 92, clocks 76, minute columns 74. Everything else goes to the three location columns.
- **The minute inputs became `width: 100%` of their cell.** A 72px input inside a 74px column is what made those two columns immovable — the input was setting the floor.
- **`.rp-leg-place { max-width: 0 }`** is the idiom that lets a fixed-layout cell shrink below its content. Without it the text columns refuse to give ground and the table overflows anyway.
- **AC2:** a stop serving several shifts sends them comma-joined, and printing all of them is what pushed the table wide in the first place. Cells now show the first place plus a `+N more` badge, with the full list on the tooltip. A single-location cell gets no badge — `+0 more` on every ordinary stop would be noise.

## Tests

`one_fm/tests/test_trip_builder_layout.py` — 18 tests, including one asserting the split separator matches the `", ".join(shift_places)` the server actually produces, and one that fails if `table-responsive` returns to this table.

| module | result |
|---|---|
| `test_trip_builder_layout` | 18 passed |
| `test_merge_preview` / `test_stop_sequence_controls` / `test_lane_day_rebase` | OK |
| `test_transportation_schedule` / `test_default_leg_timings` / `test_drop_time_departure` / `test_pool_filters` / `test_mixed_trip_merge` | OK |

**No data written** — the modal was opened via *Edit Trip Timings* and closed without confirming.


yusuffGoodie:
## Fix: a truncated header now says what it is

Reported in testing by @y-oyedele — hovering **"Target…"** showed nothing.

The leg table fits by **width, not by label** (that's what AC3 asks for), so a 76px column holding "Target Arrival" **always** ellipsizes — and so do `Departure`, `Buffer (min)`, `Transit (min)` and `On Board`.

The location **cells** have carried their full text on a tooltip since AC2. The **headers** never did. So the one place the ellipsis is guaranteed to bite was the one place with nothing behind it.

Every header now carries its own label as a `title` — the same answer AC2 already gives for the cells: the ellipsis is the layout working, not a reason to hide what the column is.

### One list, not eleven hand-written cells

```js
const headers = [
    ['Stop', 'rp-leg-num-col'],
    ...
    ['Target Arrival', 'rp-leg-time-col'],
].map(([label, cls]) => {
    const text = esc(__(label));
    return `<th${cls ? ` class="${cls}"` : ''} title="${text}">${text}</th>`;
}).join('');
```

Written out twice, the title and the label drift the first time somebody renames a column — and **a tooltip that disagrees with its header is worse than none**. One variable feeds both.

The width classes are unchanged; they have to stay on the header cell or the first data row decides the column width instead.

### Verification

**26 tests** (8 new). The rendered row was checked **under node**, not grepped:

```
<th class="rp-leg-num-col" title="Stop">Stop</th>
<th title="Accommodation / Stop">Accommodation / Stop</th>
...
<th class="rp-leg-time-col" title="Target Arrival">Target Arrival</th>
<th class="rp-leg-time-col" title="On Board">On Board</th>
```

All eleven come out with a `title` matching their text and their original class. Also pinned: every column present, in order; the label escaped once and reused; no hand-written `<th>` left (one would have no tooltip and nothing would say so); and the AC2 cell tooltips untouched.

All **343 tests** across the Transportation epic green on the rebased tip.

> Stack rebased: #6940 and #6941 force-pushed. All remain `mergeable: true`.

NoBoneZ (approved):

PR #6941, opened 2026-09-17
WI-002545: the trip duration check found a real defect

yusuffGoodie:
> **Stacked:** 6930 → … → [#6940](https://github.com/ONE-F-M/one_fm/pull/6940) → this. **Final item of the Transportation epic.**

## AC1 asked me to check the calculation. It was wrong.

AC1 says the Active Trip duration *"needs to be checked against multiple vehicles… to confirm the calculation is triggering as per the logic"*. Checking it across six vehicles turned up this:

| vehicle | Total Time | Trip Time | |
|---|---|---|---|
| VHL-L-0004 | 9h 31m | 6h 28m | ok |
| VHL-L-0006 | 16h 5m | 6h 52m | ok |
| **VHL-L-0010** | **1h 40m** | **8h 7m** | ⚠️ eight hours of driving inside a 100-minute shift |
| **VHL-L-0007** | **3h 55m** | 3h 30m | ⚠️ understated |

**Root cause — the same trap [#6933](https://github.com/ONE-F-M/one_fm/pull/6933) documented, on the server side.** The route's two ends were picked with `min()`/`max()` over the **ISO strings**. Those stamps carry two things: the TIME is the daily trip window, the DATE is the multi-day lock's lifespan (TR-8). Sorting them whole sorts by the **lock**, not by when the bus runs — so a vehicle whose rows carry different lock dates had its span measured between two unrelated days and then reduced `% 86400` into an arbitrary remainder.

**After:**

| vehicle | before | after |
|---|---|---|
| VHL-L-0010 | `1h 40m / 8h 7m` | **`19h 23m / 8h 7m`** (badge `04:00 → 23:23`) |
| VHL-L-0007 | `3h 55m / 3h 30m` | **`14h 55m / 3h 30m`** (badge `05:05 → 20:00`) |

Both ends are now chosen by clock time, the span wraps at midnight, and the per-row durations go through the same rule rather than taking a possibly-negative difference modulo a day. **All eight vehicles checked now report Total ≥ Trip** (`incoherent: []`).

## AC2–AC5: the navigation around that number

- **AC2** — the TRIP TIME badge carries an itemised breakdown: `S-101: 04:50 → 05:59 (1h 9m)` per run. Native `title`, so it works on hover *and* tap with no popover library.
- **AC3** — a sticky pill per run under the header; clicking smooth-scrolls to that trip block. The handler is **delegated from the container**, because bound to the pills it would die with the first tab change.
- **AC4/AC5** — the shift as one bar: solid where the bus is on a run, light where it's standing, each segment naming itself on hover in AC5's exact wording (`Driving Segment: … | … → …` / `Waiting Segment: Rest Time | … → …`).

All three are measured from **one pass** over the trips. A run that reads 2h05m in the tooltip and draws a different width on the strip is worse than not drawing it. The strip is positioned in **percentages** so it's correct on first paint and stays correct on resize, and a single-stop run keeps a minimum width so it doesn't vanish.

## Tests

`one_fm/tests/test_manifest_trip_duration.py` — 22 tests, including the exact stamp pair that caused the defect (clock order opposite to date order) and an exhaustive sweep asserting no gap can reach a full day.

| module | result |
|---|---|
| `test_manifest_trip_duration` | 22 passed |
| `test_manifest_vehicle_tabs` / `test_manifest_timestamp_math` / `test_transportation_manifest_page` | OK |
| `test_trip_isolated_attendance` / `test_transportation_schedule` / `test_forward_departure_scheduling` | OK |
| `test_lane_day_rebase` / `test_merge_preview` / `test_roster_overlay` | OK |


yusuffGoodie:
## Added: a capacity overload blocks its own run, not the whole bus

Reported in testing by @y-oyedele: editing **S-703's timings** was refused with

> *Capacity Exceeded: the outbound run on VHL-L-0013 carries 26 passengers but the vehicle takes 22.*

**The overload is real** — `TS-0599` genuinely puts 26 people on a 22-seat bus. But it is a **different run on the same vehicle**, already saved that way. Nothing in that save had touched it and nothing in that save could fix it, so the dispatcher was refused an unrelated edit and told to go find a larger bus for a run they weren't looking at.

### What changed

`_vehicles_this_save_did_not_touch` already grandfathers a bus nobody went near. It was **too coarse**: editing one run's timings puts its vehicle in play, and every other run on that vehicle was then judged too.

The same membership rule now also answers at **trip** granularity — a run whose cards nobody added or removed keeps the verdict it was saved with. The lane still colours it, and the moment somebody drops a card on that run it is judged in full.

The message now **names the run**. "The outbound run on VHL-L-0013" describes a bus that may hold five of them, and the one at fault is not the one being edited:

```
Capacity Exceeded: the outbound run S-703 on VHL-L-0013 carries 26 passengers
but the vehicle takes 22. You are short 4 seat(s) — ...
```

An unnamed run still reads as a sentence rather than `run None`.

The **concurrent-overlap** check is deliberately left at vehicle granularity: two trips that overlap in time genuinely *do* share the bus, so neither one alone is at fault.

### Verification

**92 route plan tests**, 7 new — the run that gained a card is judged while the overloaded one beside it is not; a brand-new plan grandfathers nothing; the message carries the name, the counts and the shortfall.

Still green: `test_overcapacity_card_split` (25), `test_route_plan_assignment_mixed` (18), `test_vehicle_passenger_capacity` (13).

### Why it is on this branch

It touches only `route_plan.py`, belongs to no single work item, and sits at the top of the stack — putting it lower would mean re-basing eight branches for a fix none of them depend on.

yusuffGoodie:
## Fix: Trip Time is each run end to end, not the sum of its rows

Reported by @y-oyedele on **60/59220**: the header read **12h 22m** where the runs add to **12h 03m**.

### What it was doing

```python
trip_ms = sum(_clock_gap(r.start_time, r.end_time) for r in v_rows)
```

Summing the assignment **rows**. That counts a shared stop twice — a merged run sets down and picks up at the same place in the same minute, so it holds two rows over one window:

```
stop=2 Alghanim  02:15 -> 02:32   (Outward card)
stop=2 Alghanim  02:15 -> 02:32   (Return card)   ← the same 17 minutes, again
```

Row-summing **also dropped the camp and home legs**, so the figure was wrong in both directions at once and only looked plausible because the errors partly cancelled.

| VHL-L-0004 | |
|---|---|
| sum of its rows | **12h 22m** ← what the header showed |
| its seven runs, departure → final arrival | **12h 03m** |

### Why it matters beyond the number

The header disagreed with the **per-trip breakdown printed directly underneath it**. AC2's popover already measures each run end to end; the two now come from one definition, so the tooltip and the badge can't contradict each other.

### Verified across all 25 vehicles

AC1 asks for exactly this — *"checked against multiple vehicles"*:

| vehicle | trips | Trip Time | Total |
|---|---|---|---|
| **VHL-L-0004** | 7 | **12h 03m** | 17h 51m |
| VHL-L-0006 | 6 | 9h 05m | 17h 00m |
| VHL-L-0010 | 8 | 8h 26m | 19h 31m |
| VHL-L-0013 | 5 | 5h 08m | 17h 20m |

**No vehicle's Trip Time exceeds its Total Time** — eight hours of driving inside a two-hour shift is what WI-002614 was reported for, and that invariant now has a test of its own.

### Verification

**32 tests**, 10 new: a shared stop counted once, camp and home legs inside the run, separate runs kept separate, the parked gap between them excluded, an ungrouped row as its own run, a run crossing midnight, a row with no start skipped rather than zeroed, and the reported vehicle's seven runs adding to exactly 12h 03m.

Related suites green: `test_manifest_timestamp_math` (39), `test_merge_preview` (97), `test_manifest_vehicle_tabs` (19), `test_trip_isolated_attendance` (19), manifest page (26).

NoBoneZ (approved):

PR #6958, opened 2026-09-20
WI-002094/WI-002091: Excess Nationals, and who is on the licence [version-15]

yusuffGoodie:
## version-15 port of #6957 — this is the branch production runs

> *"Formulas are not working, why excess is not showing, it is simple Subtraction"* · *"Instead it is showing Negative count"* · *"There is issue in counting number of employees as well"*

Full diagnosis and the projected figures are in **#6957**. This describes only what is different here.

### Short version

**#6927 computes nothing wrong.** It ran on production at `2026-09-20 08:09:49` and every figure it wrote was verified against the implemented formulas afterwards — 0 mismatches, nothing negative, nothing blank, and no client script or server script on these doctypes to render anything else. What it did was remove the zeros that had been hiding two defects since August:

1. **Excess Nationals subtracted the wrong way round.** `required − actual` is a shortfall; the clamp *"if negative, display 0"* then hides every real excess, which is the whole of what `exceeding_the_ratio_number_of_national_workers` is named after. 5 of the 12 live rows showed 0 where the licence genuinely had surplus nationals. Now `actual − required`, matching the call already made on the violation line.
2. **The headcount counted only Active.** 2,645 employees carry the two live licence numbers and 1,463 were counted; Left (1,092) is right to exclude, but Vacation (63), Not Returned from Leave (22), Court Case (4) and Absconding (1) are still employed under the licence. Now everybody but Left.

Both confirmed with the process owner.

### What is different on this branch

The three source files applied **cleanly** — `derived_figures` and `count_workers` are byte-identical across the two branches, which was verified before the port rather than assumed.

`patches.txt` conflicted, and the resolution matters: the cherry-pick wanted to bring `delete_pathfinder_log` across with it. **That patch does not exist on version-15**, and adding its line here would have failed `bench migrate` on production. Only `restate_pam_license_sector_figures` is added.

`test_pam_license_worker_counts` runs 13 here against 18 on staging — the difference is the added-sector-row and insert-path cases that are not on this branch, unrelated to this change.

### Tests

Run on this branch, against this branch's controller:

```
one_fm.grd.doctype.pam_license_details.test_pam_sector_figures          14  OK
one_fm.grd.doctype.pam_license_details.test_pam_license_worker_counts   13  OK
one_fm.grd.doctype.pam_license_details.test_pam_expats_allowed          13  OK
one_fm.grd.doctype.pam_license_details.test_pam_compliance_status        8  OK
one_fm.grd.doctype.pam_license_details.test_pam_license_configuration    8  OK
one_fm.grd.doctype.pam_license_details.test_pam_file_link_migration      5  OK
one_fm.tests.test_backfill_pam_license_sector_headcounts                10  OK
```

### Testing notes

`bench migrate`, then `bench restart`.

Expect **`مقدمي خدمات` to flip to Non-Compliant on both licences** — over its expat allowance by 28 and 12 once the still-employed staff are counted. That is the true position, not a regression, but it is a compliance status the business will see change.

Also worth a separate look: **375 Active employees, 26% of the licence population, now sit in the exempt sector** `مهن غير مشمولة بالنسبة`, which is forced to 0 allowed and 0 violated. 181 designations were mapped there by hand after the backfill ran. If any should not be exempt, the compliance picture is off in a way no code change here will fix.


one_fm/grd/doctype/pam_license_details/pam_license_details.py
```diff
@@ -130,8 +134,16 @@ def derived_figures(sector, ratio, nationals, expatriates):
 	that sector. All three come out as no requirement rather than as an error - the ratio is
 	typed by hand and a licence should not refuse to save because one row is unfilled.
 
-	Excess Nationals is what the licence is still short of that requirement, and never less
-	than zero: a sector already carrying enough nationals is not short of any.
+	Excess Nationals is how many nationals the licence carries over that requirement, and
+	never less than zero: a sector short of the requirement is not in excess of it.
+
+	WI-002094 writes that subtraction the other way round - required minus actual - which
```
Talleyrand333, line 140: @yusuffGoodie  Please Update the agent to remove work item references in code. also based on the latest coding standard we need to keep the commenting to facts only

Talleyrand333 (approved):

